PR #4, opened 2019-05-22
DOCSP-5634: attach an id to each heading

i80and:


i80and:
@skerschb We've decided to go ahead with this PR as-is

snooty/gizaparser/nodes.py
```diff
@@ -283,8 +285,14 @@ def render_heading(self, parse_rst: EmbeddedRstParser) -> Sequence[SerializableT
             heading_text = 'Optional: ' + heading_text
 
         result = parse_rst(heading_text, self.line, True)
+
+        # Generate an anchor ID for this heading
+        heading_id = docutils.nodes.make_id(
```
skerschb: what are the uniqueness guarantees of this ID? If none, mention in comments.

PR #6, opened 2019-06-02
DOCSP-5368: Avoid unnecessarily reprocessing assets

i80and:
This involves substantial restructuring:

* DependencyGraph is replaced with networkx's directed graph class.

* FileId is introduced: it's a subclass of PurePosixPath, and uniquely and
  unambiguously refers to a page. Replaces a lot of instances of str.

* Page.get_id() is replaced with Page.fake_full_path() to more accurately
  capture what it does.

* PendingTasks are introduced: instead of resolving literalinclude and figure
  immediately, they are resolved after the AST is built in a worker process.

  These pending tasks are applied by Page.finish().

  This allows for...

* The project Cache, which correlates a (FileId, int) pair with an arbitrary
  object and a version integer. The version integer is incremented each time
  a given key is updated, and is used for our tests. The integer can be used
  for cases where multiple cached entries may be required for the same FileId.
  For example, literalinclude.

  Resolving these on the main process allows the central cache to be updated,
  rather than workers' copies of said cache.

* Project is turned into a thread-safe wrapper around the underlying _Project
  class.

* A file dependency graph (_Project._expensive_operation_cache) is introduced,
  allowing us to track which files depend on which static assets.

* A filesystem monitor is added to Project. When a static asset is changed,
  the dependent source files are re-parsed.

* Tests are added, bringing us up to 88% line coverage.


i80and:
This commit is FAR larger and more sprawling than it should be. The retro on this ticket should be fun

i80and:
@skerschb Can you give an example of where you want additional comments, just so I know your expectations?

skerschb:
As a rule every file should have comments at the top explaining how it fits
into the overall scheme (what is the functional purpose) and every public
method should have a one line description of what it does and what the
parameters are.

I realize that python has unenforced encapsulation. Private methods that
are obtuse should also have descriptions.

In other words try to put yourself in Raymond’s shoes when he picks up this
source code and comment accordingly.

PR #7, opened 2019-06-24
DOCSP-5424: Validate include files within repo

rayangler:


snooty/test_parser.py
```diff
@@ -178,6 +178,35 @@ def test_literalinclude() -> None:
...
+    page, diagnostics = parse_rst(
+        parser,
+        path,
+        """
+        .. include:: /driver-examples/fake-include.rst
+        """,
+    )
+    page.finish(diagnostics)
+    assert len(diagnostics) == 1
+
```
i80and: We should test a generated path as well

snooty/parser.py
```diff
@@ -347,6 +347,29 @@ def handle_directive(
...
+                    )
+                )
+                return
+
+            fileid, path = util.reroot_path(
+                Path(argument_text), self.docpath, self.source_path
+            )
+
+            # Validate if file exists
+            if not path.exists():
```
i80and: `.is_file()` would be preferable to `.exists()`; the latter would permit a directory to be provided, for example

snooty/parser.py
```diff
@@ -347,6 +347,29 @@ def handle_directive(
                 msg = f'Invalid "literalinclude": {err}'
                 self.diagnostics.append(Diagnostic.error(msg, util.get_line(node)))
             return
+        elif name == "include":
+            if argument_text is None:
```
i80and: Ideally this would be validated by the parser. The fact that we don't currently support that is a source of great personal shame

PR #8, opened 2019-06-25
DOCSP-5704: Reject YAML with duplicate refs

rayangler:


snooty/gizaparser/nodes.py
```diff
@@ -203,7 +205,9 @@ def add(self, path: Path, text: str, elements: Sequence[_I]) -> None:
 
             self.dg.add_edge(file_id, inherit.file)
 
-    def reify(self, obj: _I, diagnostics: List[Diagnostic]) -> _I:
+    def reify(
+        self, obj: _I, diagnostics: List[Diagnostic], refs_set: Set[Optional[str]]
```
i80and: Why is the refs set a set of optional strings? I'm not sure it makes sense to put None into a set.

snooty/gizaparser/nodes.py
```diff
@@ -240,28 +244,44 @@ def reify(self, obj: _I, diagnostics: List[Diagnostic]) -> _I:
...
+        ]
+
+        return dataclasses.replace(node, data=data)
 
     def reify_all_files(
         self, diagnostics: Dict[PurePath, List[Diagnostic]]
     ) -> Iterator[Tuple[str, GizaFile[_I]]]:
         """Resolve inheritance and substitution in all source files within this category."""
+
+        refs_dict: Dict[str, Set[Optional[str]]] = dict()
```
i80and: See above about the optionality of the set.

Trivial nit, but I'd just use `{}` instead of `dict()`. It doesn't REALLY matter, but the former just involves a `BUILD_MAP` opcode, while the latter requires `LOAD_GLOBAL` followed by `CALL_FUNCTION`, which is a pretty slow double-whammy.

snooty/gizaparser/nodes.py
```diff
@@ -240,28 +244,44 @@ def reify(self, obj: _I, diagnostics: List[Diagnostic]) -> _I:
             obj.ref = ""
 
         obj = inherit(self.project_config, obj, parent, diagnostics)
+
+        # Check if ref already exists within the same file
+        if obj.ref in refs_set:
+            msg = f"ref {obj.ref} already exists"
+            diagnostics.append(Diagnostic.error(msg, obj.line))
+        else:
+            refs_set.add(obj.ref)
```
i80and: ...specifically, we don't need to add it if it's None

PR #15, opened 2019-07-31
DOCSP-6277: Add rerooting doc role target path to language server

rayangler:


snooty/language_server.py
```diff
@@ -268,15 +268,21 @@ def m_initialized(self, **kwargs: object) -> None:
...
+    def m_text_document__resolve(self, filename: str, docpath: str, resolve_type: str) -> str:
         """Given an artifact's path relative to the project's source directory,
         return a corresponding source file path relative to the project's root."""
 
         if self.project is None:
             logger.warn("Project uninitialized")
             return path
 
-        return str(self.project.config.source_path) + path
+        if resolve_type == "doc":
```
i80and: Can we update `test_text_doc_resolve()` in `test_language_server.py`?

RPC-methods.md
```diff
@@ -0,0 +1,23 @@
+# RPC Methods
+
+`language_server.py` utilizes RPC methods that the snooty parser uses to communicate with the VS Code extension.
+
+## resolve
```
i80and: This is actually `textDocument/resolve`

RPC-methods.md
```diff
@@ -0,0 +1,23 @@
...
+### Parameters
+
+| Parameter           | Description                                              |
+| :------------------ | :------------------------------------------------------- |
+|`fileName`: `str`    | Name of the target or directive to be resolved.          |
+|`docPath`: `str`     | Path of the text document that `fileName` is found in.   |
+|`resolveType`: `str` | Identifies how the method should resolve the `fileName`. |
+
+Resolve types that are currently supported:
+* `directive`
```
i80and: What is the behavior of each of these enum values?

snooty/test_language_server.py
```diff
@@ -82,11 +82,24 @@ def test_text_doc_resolve() -> None:
...
             == str(server.project.config.source_path) + test_file
         )
+
+        # Resolve arguments for testing doc role target file
+        test_file = "index"  # Testing relative path for example
+        resolveType = "doc"
+
+        assert (
+            server.m_text_document__resolve(test_file, docpath_str, resolveType)
+            == str(server.project.config.source_path) + "/" + test_file + ".txt"
```
i80and: Not that we support Windows right now, but leaving it open for future support, path comparisons like this should use Path rather than string concatenation

Note that the `resolve()` handler returns `str(path)`, so on Windows `\` path separators will be used.

snooty/language_server.py
```diff
@@ -268,15 +268,27 @@ def m_initialized(self, **kwargs: object) -> None:
...
+        if resolveType == "doc":
+            target = PurePath(fileName).with_suffix(".txt")
+            fileid, target_path = util.reroot_path(
+                target, PurePath(docPath), self.project.config.source_path
+            )
+            return str(target_path)
+        elif resolveType == "directive":
+            return str(self.project.config.source_path) + fileName
+        else:
+            logger.warn("resolveType is not supported")
```
i80and: This should be an error

PR #17, opened 2019-08-13
DOCSP-6493: Add missing tabid option to tab directive

i80and:
With this change, tab nodes will provide the tabid in the `tabid` option, and the title if given in a directive_argument.

This is opposed to the old AST format where the argument could contain either a title or a tabid depending on whether the tabset was anonymous or not.


i80and:
@sophstad Ready for another look! Sorry about the delay

PR #18, opened 2019-08-18
DOCSP-6598: Implement check_ast_testing_string() for nicer testing

i80and:


snooty/test_parser.py
```diff
@@ -123,24 +121,27 @@ def test_literalinclude() -> None:
...
-            '     "size": {"h": 27.9, "w": 35.5, "uom": "cm"}},\n\n',
-            '    {"item": "mousepad",\n',
-            '     "qty": 25,\n',
-            '     "tags": ["gel", "blue"],\n',
-            '     "size": {"h": 19, "w": 22.85, "uom": "cm"}}])</code>',
-            "</root>",
+    check_ast_testing_string(
+        page.ast,
+        "<root>"
+        + "".join(
```
rayangler: Why does this particular string require `.join()`? Is this something that is necessary with `<code>`?

snooty/gizaparser/test_extracts.py
```diff
@@ -3,7 +3,7 @@
 from .extracts import GizaExtractsCategory
 from ..types import Diagnostic, Page, EmbeddedRstParser, ProjectConfig
 from ..parser import make_embedded_rst_parser
-from ..util import ast_to_testing_string
+from ..util_test import ast_to_testing_string
```
rayangler: Is there a reason why we're using `ast_to_testing_string` here and not `check_ast_string`?

snooty/util_test.py
```diff
@@ -0,0 +1,91 @@
...
+
+def assert_etree_equals(e1: ET.Element, goal: ET.Element) -> None:
+    """Assert that two XML Elements are the same. If there is a difference in a child,
+       log the difference to stderr."""
+    assert e1.tag == goal.tag
+    if e1.text and goal.text:
+        assert (e1.text.strip() if e1.text else "") == (
+            goal.text.strip() if goal.text else ""
+        )
+    if e1.tail and goal.tail:
```
rayangler: Do we need to take into account if one has a tail and another doesn't?

snooty/gizaparser/test_extracts.py
i80and: Mainly there's a lot of tests, and I didn't want to go through and redo all of them when they're not broken, since that would be a significant time sink

snooty/util_test.py
i80and: Actually, yes, good point!

i80and: Updated!

snooty/test_parser.py
i80and: Updated!

PR #23, opened 2019-08-30
DOCSP-6442: Parse repo-wide substitutions

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-6442)] Parse substitutions defined in the `snooty.toml` under `[substitutions]` and include their nodes in the AST.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -8,7 +8,7 @@ import subprocess
 import threading
 from functools import partial
 from pathlib import Path, PurePath
-from typing import Any, Dict, Tuple, Optional, Set, List, Iterable
+from typing import cast, Any, Dict, Tuple, Optional, Set, List, Iterable
 from typing_extensions import Protocol
 import docutils.utils
 import watchdog.events
@@ -172,9 +172,12 @@ class JSONVisitor:
     """Node visitor that creates a JSON-serializable structure."""
 
     def __init__(
-        self, source_path: Path, docpath: PurePath, document: docutils.nodes.document
+        self,
+        project_config: ProjectConfig,
+        docpath: PurePath,
+        document: docutils.nodes.document,
     ) -> None:
-        self.source_path = source_path
+        self.project_config = project_config
         self.docpath = docpath
         self.document = document
         self.state: List[Dict[str, Any]] = []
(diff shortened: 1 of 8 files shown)


snooty/parser.py
```diff
@@ -536,6 +552,16 @@ def __init__(
...
         }
 
+        substitution_nodes: Dict[str, SerializableType] = {}
+        for k, v in self.config.substitutions.items():
+            node, diagnostics = parse_rst(self.parser, root, v)
+            substitution_nodes[k] = node.ast
+
+        self.config.substitution_nodes = substitution_nodes
+
+        # Re-parse using our updated config that now includes substitutions represented as nodes
```
i80and: Is this actually required?

snooty/types.py
```diff
@@ -260,6 +260,8 @@ class ProjectConfig:
     name: str
     source: str = field(default="source")
     constants: Dict[str, object] = field(default_factory=dict)
+    substitutions: Dict[str, str] = field(default_factory=dict)
+    substitution_nodes: Dict[str, SerializableType] = field(default_factory=dict)
```
i80and: I'd have a comment on this attribute noting that it's created after parse time

snooty/test_parser.py
```diff
@@ -244,6 +244,65 @@ def test_admonition() -> None:
...
+|yaml| ain't markup language.
+""",
+    )
+    assert diagnostics == []
+    check_ast_testing_string(
+        page.ast,
+        """<root>
+        <paragraph>
+        <substitution_reference>
+        <paragraph>
```
i80and: I don't think the substituted text should be in its own paragraph; we should peel that off as well

snooty/types.py
```diff
@@ -260,6 +260,9 @@ class ProjectConfig:
     name: str
     source: str = field(default="source")
     constants: Dict[str, object] = field(default_factory=dict)
+    substitutions: Dict[str, str] = field(default_factory=dict)
+    # This variable is populated after parse time with the nodes from `substitutions` strings
```
i80and: Well, more accurately, it's more along the lines of
```
substitution_nodes contains a parsed representation of the substitutions member, and is populated on Project initialization.
```

snooty/test_parser.py
```diff
@@ -244,6 +244,63 @@ def test_admonition() -> None:
...
+        """
+|yaml| ain't markup language.
+""",
+    )
+    assert diagnostics == []
+    check_ast_testing_string(
+        page.ast,
+        """<root>
+        <paragraph>
+        <substitution_reference>
```
i80and: The `<substitution_reference>` should be gone too, shouldn't it?

sophstad: At the moment, we're taking the child nodes returned by `parse_rst()` and assigning them to the `children` property of the substitution reference node. With this approach, I believe the `<substitution_reference>` tag should still be present. I also think that keeping it in the AST can help with debugging on the front-end. But if you feel that it isn't logical to include it in the AST, I'm open to that approach as well!

PR #24, opened 2019-09-04
DOCSP-6504: Fix parsing of versionadded, versionchanged, and deprecated directives

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-6504)] Both the numerical "version" argument (which appears on the same line as the directive) and the "explanation", which appears on the following line, are output as members of a node's `argument` array. Any subsequent indented directives are output as the node's children. I've already tested and verified that this output works on the front-end.

### DIFF
--- a/snooty/rstparser.py
+++ b/snooty/rstparser.py
@@ -365,6 +365,38 @@ class CodeDirective(docutils.parsers.rst.Directive):
         return [node]
 
 
+class VersionDirective(docutils.parsers.rst.Directive):
+    required_arguments = 1
+    optional_arguments = 1
+    has_content = True
+    final_argument_whitespace = True
+    option_spec: Dict[str, object] = {}
+
+    def run(self) -> List[docutils.nodes.Node]:
+        source, line = self.state_machine.get_source_and_line(self.lineno)
+        node = directive(self.name)
+        node.document = self.state.document
+        node.source, node.line = source, line
+        node["options"] = self.options
+
+        if self.arguments is not None:
+            textnodes = []
+            for argument_text in self.arguments:
+                text, messages = self.state.inline_text(argument_text, self.lineno)
+                textnodes.extend(text)
+            argument = directive_argument("", "", *textnodes)
+            argument.document = self.state.document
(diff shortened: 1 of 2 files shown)


snooty/test_parser.py
```diff
@@ -615,3 +615,97 @@ def test_cond() -> None:
...
+        1.8
+        </text>
+        </directive>
+        <paragraph>
+        <text>A new paragraph.</text>
+        </paragraph>
+        </root>""",
+    )
+
+    page, diagnostics = parse_rst(
```
i80and: What's the difference between this and the first test?

snooty/test_parser.py
```diff
@@ -615,3 +615,97 @@ def test_cond() -> None:
         </directive>
         </root>""",
     )
+
+
+def test_version() -> None:
```
i80and: I'd actually name this `test_versionchange`

snooty/rstparser.py
```diff
@@ -365,6 +365,38 @@ def run(self) -> List[docutils.nodes.Node]:
         return [node]
 
 
+class VersionDirective(docutils.parsers.rst.Directive):
```
i80and: I'm a little hypocritical since I'm not the best role model in this regard, but a docstring here would be good describing what the directive is, and why we need to override the normal directive handling behavior

snooty/test_parser.py
sophstad: I wanted to have tests for all directives: `versionadded`, `versionchanged`, and `deprecated`. But perhaps it is redundant?

snooty/test_parser.py
sophstad: See above, I named it `test_version` because it tests 3 different directives. But they are all functionally the same so I could see it making sense to use `test_versionchange` instead!

snooty/test_parser.py
i80and: It's all the same implementation, so it's really not necessary. But I don't think it's worth doing another round for :+1:

PR #26, opened 2019-09-06
DOCSP-6805: Doc links with dotted filenames are validated properly

rayangler:
[Jira ticket](https://jira.mongodb.org/browse/DOCSP-6805)

I refactored the process of converting the doc target name into its own util function `add_doc_target_ext()` to keep the logic consistent between the parser's doc target validation and the language server

### DIFF
--- a/snooty/test_util.py
+++ b/snooty/test_util.py
@@ -40,6 +40,33 @@ def test_get_files() -> None:
     }
 
 
+def test_add_doc_target_ext() -> None:
+    # Set up target filenames
+    root = Path("root")
+    docpath = PurePath("path/to/doc/")
+    targets = [
+        "dottedfilename",
+        "dotted.filename",
+        "dotted.file.name",
+        "d.o.t.t.e.d.f.i.l.e.n.a.m.e",
+    ]
+
+    # What we want the resulting target paths to be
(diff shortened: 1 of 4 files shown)


rayangler:
Just added a test in `test_util.py`. I decided to put the test here to avoid creating multiple files in `test_project` and because testing the core util function will be beneficial for both the parser and the language server.

snooty/test_util.py
```diff
@@ -40,6 +40,34 @@ def test_get_files() -> None:
...
+        path_to_file.joinpath(Path("dotted.filename.txt")).resolve(),
+        path_to_file.joinpath(Path("dotted.file.name.txt")).resolve(),
+        path_to_file.joinpath(Path("d.o.t.t.e.d.f.i.l.e.n.a.m.e.txt")).resolve(),
+    ]
+
+    # Compare resulting target paths from add_doc_target_ext()
+    test_results = [
+        util.add_doc_target_ext(target, docpath, root) for target in targets
+    ]
+    for i in range(len(test_results)):
```
i80and: This would be more idiomatic as either
```
for expected, actual in zip(resolved_targets, test_results):
    assert expected == actual
```
or just
```
assert resolved_targets == test_results
```
I'd be okay with either!

PR #29, opened 2019-09-09
DOCSP-6447: Parse card-group directive

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-6447)] Parse card group and upload associated images to `snooty.assets` collection.

### DIFF
--- a/snooty/rstparser.py
+++ b/snooty/rstparser.py
@@ -56,6 +56,15 @@ class LegacyTabsDefinition(nodes.Node):
 @checked
 @dataclass
 class CardDefinition(nodes.Node):
+    """Represents a Card within a CardGroup.
+
+    Attributes:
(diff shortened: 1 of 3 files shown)


snooty/rstparser.py
```diff
@@ -240,6 +266,61 @@ def prepare_viewlist(text: str, ignore: int = 1) -> List[str]:
     return lines
 
 
+class CardGroupDirective(BaseDocutilsDirective):
+    required_arguments = 0
+    optional_arguments = 0
+    has_content = True
+    option_spec = {
+        "type": lambda type: docutils.parsers.rst.directives.choice(
```
i80and: This is a little picky, but `type` here is shadowing the type() Python builtin

PR #31, opened 2019-09-23
DOCSP-6848: Add automatic release makefile target

i80and:


tools/bump_version.py
```diff
@@ -0,0 +1,106 @@
...
+
+    return result
+
+
+def main() -> None:
+    version_to_bump_to = sys.argv[1] if sys.argv[1] != "dev" else None
+    if version_to_bump_to is not None:
+        if not INCOMING_VERISON_PAT.match(version_to_bump_to):
+            print(
+                f'Invalid incoming version string "${version_to_bump_to}"',
```
skerschb: Maybe give a format example to describe the proper format?

HACKING.md
```diff
@@ -49,46 +51,89 @@ The parser transforms Giza-style YAML files using the `gizaparser`
...
 This will generate an HTML representation of code coverage throughout the repo that can be viewed in the browser.
 
-# Problem Areas
+### Release Process
+
+To release snooty, do the following:
+
+* Ensure that the "Unreleased" section of CHANGELOG.md is up-to-date.
+
+* Run `make cut-release BUMP_TO_VERSION="<new_version>"`, and make note of the release name that
```
sophstad: Are there any standards we want to follow for version naming? I don't have much experience with it, so perhaps it could be useful to either explain our approach or link out to another resource somewhere in this document. I think Sue's comment addresses this as well!

HACKING.md
```diff
@@ -49,46 +51,89 @@ The parser transforms Giza-style YAML files using the `gizaparser`
...
+  is printed at the end.
+
+* Ensure that everything looks okay, and that the binary generated in `dist/` works correctly.
+
+* Run:
+
+  ```shell
+  git push --tags
+  git push
```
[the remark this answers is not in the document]
i80and: `git push` won't commit tags, and `git push --tags` won't push branches. Actually, I can combine this into just `git push --follow-tags` it seems!

HACKING.md
i80and: Added more details!

PR #35, opened 2019-09-24
DOCSP-6904: Fix truncated extracts

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-6904)] The removal of suffixes from includes was causing filenames to be truncated if they contained a period (`.`). Now, don't remove suffix if the file is an extract.

### DIFF
--- a/snooty/main.py
+++ b/snooty/main.py
@@ -84,7 +84,12 @@ class MongoBackend(Backend):
             asset.get_checksum() for asset in page.static_assets if asset.can_upload()
         )
 
-        fully_qualified_pageid = "/".join(prefix + [page_id.with_suffix("").as_posix()])
+        if "extracts" in page_id.as_posix():
+            fully_qualified_pageid = "/".join(prefix + [page_id.as_posix()])
+        else:
+            fully_qualified_pageid = "/".join(
+                prefix + [page_id.with_suffix("").as_posix()]
+            )
         self.client["snooty"]["documents"].replace_one(
             {"_id": fully_qualified_pageid},
             {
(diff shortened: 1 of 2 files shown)


snooty/main.py
```diff
@@ -84,7 +84,12 @@ def on_update(self, prefix: List[str], page_id: FileId, page: Page) -> None:
             asset.get_checksum() for asset in page.static_assets if asset.can_upload()
         )
 
-        fully_qualified_pageid = "/".join(prefix + [page_id.with_suffix("").as_posix()])
+        if "extracts" in page_id.as_posix():
```
i80and: Mmmmm... this will be quite brittle I think and just masks the underlying problem: that `with_suffix("")` *cannot* deal properly with files that may or may not have an extension.

I'm not sure whether or not we even need that; will a page_id ever have an extension? Try just removing that method call, and see if any tests break. Also try adding `assert page_id.suffix not in {".txt", ".rst", ".yaml"}` and see if it barfs on building some of the cloud repos.

If just removing `with_suffix("")` turns out to not work, another approach would be to just strip `\.(txt)|(rst)|(yaml)$` from the id using `page_id.with_name()` and `page_id.name`

snooty/main.py
```diff
@@ -84,12 +85,11 @@ def on_update(self, prefix: List[str], page_id: FileId, page: Page) -> None:
             asset.get_checksum() for asset in page.static_assets if asset.can_upload()
         )
 
-        if "extracts" in page_id.as_posix():
-            fully_qualified_pageid = "/".join(prefix + [page_id.as_posix()])
-        else:
-            fully_qualified_pageid = "/".join(
-                prefix + [page_id.with_suffix("").as_posix()]
-            )
+        page_id = page_id.with_name(
```
i80and: Add a comment briefly explaining why we're doing this

PR #36, opened 2019-09-25
Make the storage engine type pluggable for testing

i80and:


i80and:
@sophstad What do you think of this patch? The idea is that `MongoBackend` can be instantiated with anything matching the pymongo interface(ish).

i80and:
The more I think about it the less I like this approach

PR #37, opened 2019-09-27
DOCSP-6175: creating test for merge conflict artifacts

madelinezec:
`test_merge_conflict.py` tests `test_data/merge_conflict` project


snooty/types.py
```diff
@@ -311,6 +312,21 @@ def render_constants(self) -> Tuple["ProjectConfig", List[Diagnostic]]:
...
+        if match:
+             diagnostics = []
+             lineno = text.count("\n", 0, match.start())
+             diagnostics.append(
+                Diagnostic.error(
+                    "merge conflict artifact found in source documentation", lineno
+                )
+             )
+             
+             return (text, diagnostics)
```
i80and: We should still run substitutions to try and pick up as many diagnostics as possible. So either way you're going to be running `self.substitute(text)`: it's just that you'll be merging the diagnostic list

snooty/test_merge_conflict.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def on_update(self, prefix: List[str], page_id: FileId, page: Page) -> None:
+        self.pages[page_id] = page
+        self.updates.append(page_id)
+
+    def on_delete(self, page_id: FileId) -> None:
+        pass
+
+
+def test() -> None:
```
i80and: Make sure that the correct diagnostic is raised on the correct line, and strip out unrelated `asserts`

snooty/test_merge_conflict.py
```diff
@@ -0,0 +1,50 @@
...
+        # Ensure that filesystem monitoring threads have been started
+        assert len(threading.enumerate()) > n_threads
+
+        # Ensure that the correct pages and assets exist
+        index_id = FileId("index.txt")
+        print("hi before keys")
+        print(backend.pages.keys())
+        assert list(backend.pages.keys()) == [index_id]
+        code_length = 0
+        checksums = []
```
i80and: Compiler is reporting an error here. Make sure you run `make format test` frequently!

snooty/test_merge_conflict.py
```diff
@@ -0,0 +1,50 @@
+import sys
```
i80and: I'd rather this test be in `test_project.py`

snooty/types.py
```diff
@@ -311,6 +312,21 @@ def render_constants(self) -> Tuple["ProjectConfig", List[Diagnostic]]:
 
     def read(self, path: Path) -> Tuple[str, List[Diagnostic]]:
         text = path.read_text(encoding="utf-8")
+        PAT_GIT_MARKER = re.compile(r"^<<<<<<< .*?^=======\n.*?^>>>>>>>", re.M | re.S)
```
i80and: This needs to be lifted out as a global constant; we don't want to be compiling a regexp every time we load a file

snooty/types.py
```diff
@@ -311,6 +312,21 @@ def render_constants(self) -> Tuple["ProjectConfig", List[Diagnostic]]:
...
+        match = PAT_GIT_MARKER.search(text)
+
+        if match:
+             diagnostics = []
+             lineno = text.count("\n", 0, match.start())
+             diagnostics.append(
+                Diagnostic.error(
+                    "merge conflict artifact found in source documentation", lineno
```
i80and: Probably instead say just `git merge conflict`

test_data/merge_conflict/snooty.toml
```diff
@@ -0,0 +1 @@
+name = "test_data"
```
i80and: `merge_conflict`

test_data/merge_conflict/source/index.txt
```diff
@@ -0,0 +1,140 @@
...
+   You must have the URL of the |onprem| Web Server, as set in the
+   You must have the |url| of the |onprem| Web Server, as set in the
+   ``mmsBaseUrl`` setting of the :doc:`{+magent+} configuration
+   file </reference/monitoring-agent>`.
+
+   Provision ten servers to host the components of the :term:`sharded
+   cluster`. For server requirements, see the :manual:`Production Notes
+   </administration/production-notes>` in the MongoDB manual.
+
+<<<<<<< HEAD
```
i80and: It might be useful to have another sample conflict where the before or after is empty

snooty/test_merge_conflict.py
madelinezec: Would you be ok with two `with Project(Path("test_data/test_project"), backend) as project: [...]` blocks inside test() in `test_project.py`

snooty/types.py
```diff
@@ -311,7 +312,15 @@ def render_constants(self) -> Tuple["ProjectConfig", List[Diagnostic]]:
 
     def read(self, path: Path) -> Tuple[str, List[Diagnostic]]:
         text = path.read_text(encoding="utf-8")
-        return self.substitute(text)
+        sourceText, diagnostics = self.substitute(text)
```
i80and: Python convention is to use snake case: `source_text`

snooty/types.py
```diff
@@ -311,7 +312,15 @@ def render_constants(self) -> Tuple["ProjectConfig", List[Diagnostic]]:
...
+        match_found = PAT_GIT_MARKER.finditer(text)
+
+        if match_found:
+            for match in match_found:
+                lineno = text.count("\n", 0, match.start())
```
i80and: Did you mean `source_text` here?

snooty/types.py
```diff
@@ -311,7 +312,15 @@ def render_constants(self) -> Tuple["ProjectConfig", List[Diagnostic]]:
...
     def read(self, path: Path) -> Tuple[str, List[Diagnostic]]:
         text = path.read_text(encoding="utf-8")
-        return self.substitute(text)
+        source_text, diagnostics = self.substitute(text)
+        match_found = PAT_GIT_MARKER.finditer(text)
+
+        if match_found:
+            for match in match_found:
+                lineno = source_text.count("\n", 0, match.start())
+                diagnostics.append(Diagnostic.error("git merge conflict found", lineno))
```
skerschb: Should the diagnostic error also report the file path to make it easier for the user to parse (and also thinking about standard automated log parsing tools)?

i80and: Diagnostics are always associated with a `Page` instance that stores the path

PR #39, opened 2019-10-02
DOCSP-5819: Parse toctree directives

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5819)] Properly parse toctree directives to prepare them for semantic parsing.

### DIFF
--- a/snooty/rstparser.py
+++ b/snooty/rstparser.py
@@ -21,7 +21,7 @@ from .flutter import checked, check_type, LoadError
 from . import util
 from . import specparser
 
-PAT_EXPLICIT_TILE = re.compile(
+PAT_EXPLICIT_TITLE = re.compile(
     r"^(?P<label>.+?)\s*(?<!\x00)<(?P<target>.*?)>$", re.DOTALL
 )
 PAT_WHITESPACE = re.compile(r"^\x20*")
@@ -138,7 +138,7 @@ def handle_role_explicit_title(
     content: List[object] = [],
 ) -> Tuple[List[docutils.nodes.Node], List[docutils.nodes.Node]]:
     """Handle link-like roles with a target and an optional title."""
-    match = PAT_EXPLICIT_TILE.match(text)
+    match = PAT_EXPLICIT_TITLE.match(text)
     if match:
         node = role(typ, lineno, match["label"], match["target"])
     else:
@@ -163,7 +163,7 @@ class LinkRoleHandler:
         options: Dict[str, object] = {},
         content: List[object] = [],
     ) -> Tuple[List[docutils.nodes.Node], List[docutils.nodes.Node]]:
-        match = PAT_EXPLICIT_TILE.match(text)
+        match = PAT_EXPLICIT_TITLE.match(text)
         label: Optional[str] = None
         if match:
             label, target = match["label"], match["target"]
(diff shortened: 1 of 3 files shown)


snooty/rstparser.py
```diff
@@ -485,6 +486,58 @@ def run(self) -> List[docutils.nodes.Node]:
...
+    required_arguments = 0
+    optional_arguments = 0
+    has_content = True
+    final_argument_whitespace = True
+    option_spec = {
+        "caption": str,
+        "hidden": util.option_flag,
+        "includehidden": util.option_flag,
+        "maxdepth": int,
+        "titlesonly": util.option_flag,
```
sophstad: These are the only options utilized in our docs repos, but there are a few more (defined [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#table-of-contents)). Do we want to support them all?

i80and: We have some discretion about what to support, but in general we should try to support everything!

snooty/rstparser.py
```diff
@@ -485,6 +486,61 @@ def run(self) -> List[docutils.nodes.Node]:
...
+        for child in self.content:
+            entries.append(self.make_toc_entry(source, child))
+        node["entries"] = entries
+
+        return [node]
+
+    def make_toc_entry(self, source: str, child: str) -> Dict[str, str]:
+        """Parse entry for either url or slug and optional title"""
+        entry: Dict[str, str] = {}
+        if "<" in child:
```
i80and: Does `PAT_EXPLICIT_TILE` (um, apparently I made a typo that should be fixed) do what you need here?

snooty/test_parser.py
```diff
@@ -847,3 +847,30 @@ def test_cardgroup() -> None:
         </directive>
         </root>""",
     )
+
+
+def test_toctree() -> None:
```
i80and: I'd love if we also had another test for the case of a URL with no title

sophstad: This case should fail, correct?

i80and: We would have an assertion that a correct diagnostic is assigned to the page

snooty/rstparser.py
```diff
@@ -517,14 +517,22 @@ def run(self) -> List[docutils.nodes.Node]:
...
-            entries.append(self.make_toc_entry(source, child))
+            entry, err = self.make_toc_entry(source, child)
+            if entry:
+                entries.append(entry)
+            if err:
+                errors.extend(err)
         node["entries"] = entries
 
-        return [node]
+        result: List[docutils.nodes.Node] = [node]
```
i80and: You can combine these lines:
```python
return [node, *errors]
```

snooty/rstparser.py
```diff
@@ -533,16 +541,20 @@ def make_toc_entry(source: str, child: str) -> Dict[str, str]:
         if match:
             label, target = match["label"], match["target"]
             entry["title"] = label
+        elif child.startswith("<") and child.endswith(">"):
+            # If entry is surrounded by <> tags, assume it is a URL and log an error.
+            err = "toctree nodes with URLs must include titles"
+            error_node = self.state.document.reporter.error(str(err), line=self.lineno)
```
i80and: `err` is already a string; `str(err)` is redundant

snooty/rstparser.py
```diff
@@ -533,16 +541,20 @@ def make_toc_entry(source: str, child: str) -> Dict[str, str]:
...
+            return (entry, [error_node])
```
i80and: I don't think you need these parenthesis

snooty/rstparser.py
```diff
@@ -533,16 +541,20 @@ def make_toc_entry(source: str, child: str) -> Dict[str, str]:
...
-            target = child.replace("<", "").replace(">", "")
+            target = child
 
         parsed = urllib.parse.urlparse(target)
         if parsed.scheme:
             entry["url"] = target
         else:
             entry["slug"] = target
-        return entry
+        return (entry, [])
```
i80and: ditto

i80and:
@skerschb pending your review

snooty/test_parser.py
```diff
@@ -847,3 +847,31 @@ def test_cardgroup() -> None:
...
```
skerschb: What is the plan for the toc,yaml files in the docs repo? If we have not addressed those in this test we might put a todo here. Otherwise LGTM.

sophstad: According to @i80and, we will run a one-time migration to convert those files into rST format. I believe we'll no longer support yaml, but in any case it's a separate issue from this ticket!

PR #42, opened 2019-10-09
DOCSP-7193: Parse published-branches.yaml and save to metadata collection

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-7193)] Parse `published-branches.yaml` from repository and upload as a document in the `metadata` collection.

After parsing BI Connector's [`published-branches.yaml`](https://github.com/mongodb/docs-tools/blob/master/data/bi-connector-published-branches.yaml), this document would be added to the `snooty.metadata` collection:
```json
{
  "_id": "bi-connector/sophstad/master",
  "publishedBranches": {
    "version": {
      "published": [
        "2.12",
        "2.11",
        "2.10",
        "2.9",
        "2.8",
        "2.7",
        "2.6",
        "2.5"
      ],
      "active": [
        "2.12",
        "2.11",
        "2.10",
        "2.9",
        "2.8",
        "2.7",
        "2.6",
        "2.5"
      ]
    },
    "git": {
      "branches": {
        "manual": "master",
        "published": [
          "master",
          "v2.11",
          "v2.10",
          "v2.9",
          "v2.8",
          "v2.7",
          "v2.6",
          "v2.5"
        ]
      }
    }
  }
}
```


snooty/language_server.py
```diff
@@ -159,6 +159,11 @@ def on_update(self, prefix: List[str], page_id: FileId, page: types.Page) -> Non
     def on_delete(self, page_id: FileId) -> None:
         pass
 
+    def on_published_branches(
+        self, prefix: List[str], published_branches: SerializableType
```
i80and: Should we pass `PublishedBranches` instead of the serialized form, and let the backend serialize things if it wants?

snooty/parser.py
```diff
@@ -613,6 +620,28 @@ def __init__(
         self.asset_dg: "networkx.DiGraph[FileId]" = networkx.DiGraph()
         self._expensive_operation_cache = Cache()
 
+        self.published_branches: SerializableType = self.get_parsed_branches()
+        self.backend.on_published_branches(self.prefix, self.published_branches)
+
+    def get_parsed_branches(self) -> SerializableType:
```
i80and: See above about the return type

snooty/parser.py
```diff
@@ -613,6 +620,28 @@ def __init__(
...
+                try:
+                    result = check_type(
+                        gizaparser.published_branches.PublishedBranches, data
+                    )
+                    return result.serialize()
+                except LoadError as err:
+                    raise ValueError("Error reading file: " + str(err))
+        except FileNotFoundError:
+            pass
+        except LoadError as err:
```
i80and: Probably also `yaml.YAMLError`

snooty/parser.py
```diff
@@ -613,6 +620,28 @@ def __init__(
...
+            raise ValueError("Error loading file: " + str(err))
```
i80and: f-strings are appropriate: `f"Error loading file: {err}"`

snooty/parser.py
```diff
@@ -613,6 +620,28 @@ def __init__(
...
+        return published_branches
```
i80and: `published_branches` is always `{}` at this point in the control flow, right?

sophstad: Indeed.

snooty/gizaparser/published_branches.py
```diff
@@ -0,0 +1,83 @@
+from dataclasses import dataclass
+from typing import Dict, List, Optional
+from ..flutter import checked
+from ..types import SerializableType
+
+"""Models for giza's published_branches files.
```
i80and: Docstring needs to be the first thing in a module, before imports

snooty/parser.py
```diff
@@ -613,6 +621,29 @@ def __init__(
         self.asset_dg: "networkx.DiGraph[FileId]" = networkx.DiGraph()
         self._expensive_operation_cache = Cache()
 
+        published_branches: Optional[PublishedBranches] = self.get_parsed_branches()
+        if published_branches:
+            self.published_branches: PublishedBranches = published_branches
```
i80and: We never actually use this anywhere else. Does it need to be an instance property?

snooty/parser.py
```diff
@@ -613,6 +621,29 @@ def __init__(
         self.asset_dg: "networkx.DiGraph[FileId]" = networkx.DiGraph()
         self._expensive_operation_cache = Cache()
 
+        published_branches: Optional[PublishedBranches] = self.get_parsed_branches()
```
i80and: If this raises an exception should we report a diagnostic instead of crashing? (See line 589. Also we should report whatever we found on line 608)

snooty/parser.py
sophstad: It's used in a test [here](https://github.com/mongodb/snooty-parser/pull/42/files#diff-63fe176602884ead559d986042bccc4eR115-R118)! Do you think it's worth keeping for this?

i80and: That test can use the `on_published_branches` handler I think!

PR #45, opened 2019-10-17
DOCSP-5351: Validate list-table structure

varsha804:
https://jira.mongodb.org/browse/DOCSP-5351


snooty/rstparser.py
```diff
@@ -486,6 +486,45 @@ def run(self) -> List[docutils.nodes.Node]:
         return [node]
 
 
+class ListTableDirective(docutils.parsers.rst.Directive):
```
i80and: I don't think you need this anymore; the default directive handler (as created by the `rstspec.toml`) should be okay I think? Try getting rid of this class and see what happens.

snooty/parser.py
```diff
@@ -365,7 +364,20 @@ def handle_directive(
             doc["argument"] = argument
 
         argument_text = None
+        expected_num_columns = 0
         try:
+            if name == "list-table":
+                # Calculate the expected number of columns for this list-table structure.
```
i80and: Try getting rid of the custom `ListTableDirective` directive handler below; then `widths` should be available via the `options` variable so you don't need to do all this string manipulation.

snooty/test_parser.py
```diff
@@ -711,6 +711,50 @@ def test_version() -> None:
...
+   * - Stage
+     - Description
+ 
+   * - :pipeline:`$geoNear`
+     - .. include:: /includes/extracts/geoNear-stage-toc-description.rst
+     - .. include:: /includes/extracts/geoNear-stage-index-requirement.rst
+""",
+    )
+    page.finish(diagnostics)
+    assert len(diagnostics) == 1
```
i80and: What line does this diagnostic report? Ideally it would be line 10

snooty/parser.py
```diff
@@ -481,6 +512,22 @@ def validate_doc_role(self, node: docutils.nodes.Node) -> None:
...
+        if (
+            isinstance(node, docutils.nodes.bullet_list)
+            and len(node.children) != expected_num_columns
+        ):
+            msg = (
+                f'expected "{expected_num_columns}" columns, saw "{len(node.children)}"'
+            )
+            self.diagnostics.append(Diagnostic.error(msg, util.get_line(node)))
+            return
+        for child in node.children:
```
i80and: Could you add a comment explaining this?

snooty/parser.py
```diff
@@ -394,6 +416,16 @@ def handle_directive(
             except OSError as err:
                 msg = f'"{name}" could not open "{argument_text}": {os.strerror(err.errno)}'
                 self.diagnostics.append(Diagnostic.error(msg, util.get_line(node)))
+
+        elif name == "list-table":
```
i80and: Should this be merged into the previous handling of `name == list-table`?

snooty/parser.py
```diff
@@ -481,6 +498,22 @@ def validate_doc_role(self, node: docutils.nodes.Node) -> None:
             msg = f'"{node["name"]}" could not open "{resolved_target_path}": No such file exists'
             self.diagnostics.append(Diagnostic.error(msg, util.get_line(node)))
 
+    def validate_toctree(
```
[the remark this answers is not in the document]
varsha804: Makes more sense!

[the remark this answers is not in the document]
varsha804: Sure, will change the function name!

Would it be odd to pass in `self.diagnostics` when it could just be a method of the class, though? Keeping the function as is would also resemble the other `validate` function!

snooty/rstparser.py
varsha804: Works!

snooty/test_parser.py
varsha804: It reports line 8

i80and: Line 8 is okay I think!

snooty/test_parser.py
```diff
@@ -711,6 +711,68 @@ def test_version() -> None:
...
+
+
```
sophstad: Mentioned above, but I think adding a nested list table test case is a good idea! [Here](https://github.com/mongodb/docs-compass/blob/master/source/includes/list-table-ssl-connection.rst) is an example in our source code.

snooty/parser.py
```diff
@@ -394,6 +393,21 @@ def handle_directive(
...
                 self.diagnostics.append(Diagnostic.error(msg, util.get_line(node)))
+
+        elif name == "list-table":
+            # Calculate the expected number of columns for this list-table structure.
+            expected_num_columns = 0
+            if "widths" in options:
+                expected_num_columns = len(options["widths"].split(" "))
+
+            for outer_bullet in node.children:
+                for bullet in outer_bullet.children:
```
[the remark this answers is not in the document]
i80and: Just as a note, the exact bullet characters don't matter: you can swap ` * ` and ` - ` for rows vs columns

snooty/test_parser.py
varsha804: I've added another test case for this!

snooty/parser.py
```diff
@@ -484,6 +497,27 @@ def validate_doc_role(self, node: docutils.nodes.Node) -> None:
...
+            self.diagnostics.append(
+                Diagnostic.error(msg, util.get_line(node) + len(node.children) - 1)
+            )
+            return
+
+        # unwrap the data structures holding the bullet list
+        if not isinstance(node, docutils.nodes.bullet_list):
+            for child in node.children:
+                self.validate_list_table(child, expected_num_columns)
+
```
sophstad: I really don't think we need to recurse here, and could delete this code entirely. Recursing is redundant: `validate_list_table()` will be called automatically on every `list-table` directive, and we don't need to run this validation on any other type of node that appears as a list table's child. Let me know if I can clarify this for you further, it's definitely a bit confusing!

varsha804: Oh I was actually using this recursive call to remove the containers around the first `bullet_list`, but I just realized that I could've just done it in a simpler way without recursion 😄. Adding that soon!

snooty/test_parser.py
```diff
@@ -711,6 +711,89 @@ def test_version() -> None:
...
+   :header-rows: 1
+
+   * - Stage
+     - Description
+ 
+   * - :pipeline:`$geoNear`
+     - .. include:: /includes/extracts/geoNear-stage-toc-description.rst
+      
+       + More nesting
+       + Some description
```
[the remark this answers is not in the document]
varsha804: Ah, got it!

PR #49, opened 2019-10-23
DOCSP-7243: Create semantic toctree

varsha804:
https://jira.mongodb.org/browse/DOCSP-7243


test_data/test_semantic_parser/source/page3.txt
```diff
@@ -1 +1,29 @@
...
+requirements for sharding collections for |global-write|, see 
+:ref:`global-writes-collections`.
+
+The |service| :ref:`data-explorer` supports creating sharded
+collections with specific validations for |global-write|. For complete
+documentation, see :ref:`de-shard-collection-for-global-writes`.
+
+.. class:: hidden
+
+   .. toctree::
```
sophstad: I think a test case with multiple `toctree` directives on a single page would be valuable!

snooty/semanticparser.py
```diff
@@ -54,3 +55,87 @@ def slug_title(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
...
+        ][0]
+
+        # Construct a {slug: fileid} mapping so that we can retrieve the full file name
+        # given a slug. We cannot use the with_suffix method since the type of the slug
+        # in find_toctree_nodes(...) is string rather than FileId.
+        for fileid in pages:
+            slug = fileid.without_known_suffix
+            fileid_dict[slug] = fileid
+
+        slug_title: Dict[str, Any] = cast(Dict[str, Any], self.slug_title(pages))[
```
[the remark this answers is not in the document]
varsha804: Ah yes! I actually had to implement that in the [breadcrumbs ticket ](https://github.com/mongodb/snooty-parser/pull/52) because of a bug I found.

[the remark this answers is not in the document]
varsha804: Ah yes! I actually had to implement that in the [breadcrumbs ticket](https://github.com/mongodb/snooty-parser/pull/52) because of a bug I found.

snooty/semanticparser.py
```diff
@@ -54,3 +55,87 @@ def slug_title(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
...
+
+    def toctree(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
+        fileid_dict = {}
+        toctree: Dict[str, Any] = {"toctree": {}}
+
+        starting_fileid: FileId = [
+            fileid
+            for fileid in pages.keys()
+            if str(fileid) == "contents.txt" or str(fileid) == "index.txt"
+        ][0]
```
sophstad: Would appreciate a comment explaining what's happening here!

snooty/semanticparser.py
```diff
@@ -54,3 +55,87 @@ def slug_title(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
...
+                node["children"].extend(ast["entries"])
+
+            # Recursively build the tree for each toctree node in this entries list
+            for toctree_node in node["children"]:
+                if "slug" in toctree_node:
+                    # Only recursively build the tree for internal links
+                    slug = toctree_node["slug"]
+                    if slug[0] == "/":
+                        slug = slug[1:]
+                    idx = slug.find(".")
```
sophstad: I believe what you're trying to do here is identify a link as a URL if it contains a `.`. That is a bit risky because filenames can actually include periods, which they often do when dealing with MongoDB numbered versions ([example](https://github.com/mongodb/docs/blob/master/source/includes/extracts-4.2-changes.yaml)). I had to approach this problem too, and ended up looking for the angle brackets that must enclose a URL to identify it—[see here](https://github.com/mongodb/snooty-parser/blob/master/snooty/rstparser.py#L582-L585).

Let me know if I've misunderstood what's going on here! And some comments when you update would be appreciated ☺️

snooty/test_semantic_parser.py
```diff
@@ -38,18 +39,48 @@ def on_published_branches(
...
+    )
+
+    assert len(slug_to_title) == 4
+    assert slug_to_title["index"] == "Connection Limits and Cluster Tier"
+    assert slug_to_title["page1"] == "Print this heading"
+    assert slug_to_title["page2"] == "Heading is not at the top for some reason"
+    assert slug_to_title["page3"] == ""
+
+
+def test_toctree(backend: Backend) -> None:
```
sophstad: Nice updates 👍

snooty/semanticparser.py
varsha804: I'm actually removing the slug's file extension here! I'm doing it this way instead of using Raymund's `without_known_suffix` function since by this point in the code, I don't have the appropriate `FileId` object, just the slug as a string.

Slugs and urls are added to the toctree in lines 111-112:
`if "children" not in node:`
     `node["children"] = ast["entries"]`

I'll add some comments to make this clear!

varsha804: Just realized this doesn't work because of filenames like `security-x.509.txt` 😬 I'll push a modification soon!

sophstad: Oh interesting! Is this something you ran into a lot? I think the vast majority of the time, slugs are enumerated without extensions.

In any case, this is something we will want to handle in the regular parser, not the semantic layer. Just include the slugs as-is here!

varsha804: Discussed offline: we will continue to strip file extensions from slugs in the semantic layer so that we can successfully find a slug's FileId object (currently we use a `{slug: FileId}` mapping to do so). This logic will be moved to the parse layer in [DOCSP-7595](https://jira.mongodb.org/browse/DOCSP-7595).

snooty/semanticparser.py
```diff
@@ -54,3 +58,93 @@ def slug_title(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
...
 
         return {"slugToTitle": slug_title_dict}
+
+    def toctree(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
+        fileid_dict = {}
+        toctree: Dict[str, Any] = {"toctree": {}}
+
+        # The toctree must begin at either `contents.txt` or `index.txt`.
+        # Generally, repositories will have one or the other; but, if a repo has both,
+        # a starting point will be arbitrarily chosen.
```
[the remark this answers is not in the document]
varsha804: Oops, you're right since `pages` should be alphabetically ordered! Will update the comment 👍

PR #52, opened 2019-11-01
DOCSP-7499: Page Breadcrumbs (Semantic Parse)

varsha804:
https://jira.mongodb.org/browse/DOCSP-7499


snooty/semanticparser.py
```diff
@@ -9,7 +9,9 @@ def __init__(self) -> None:
     def run(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
         # Specify which transformations should be included in semantic postprocessing
         functions: List[Callable[[Dict[FileId, Page]], Dict[str, SerializableType]]] = [
-            self.slug_title
```
skerschb: // TODO: wrap this in a function so that we might toggle what semantic parse layers are exposed based on an environment setting. [Useful for when we support additional transforms for REST retrieved content]

snooty/semanticparser.py
```diff
@@ -54,3 +56,117 @@ def slug_title(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
...
+
+
+def find_toctree_nodes(
+    fileid: FileId,
+    ast: Dict[str, Any],
+    pages: Dict[FileId, Page],
+    node: Dict[Any, Any],
+    fileid_dict: Dict[str, FileId],
+) -> None:
+    if "children" not in ast.keys():
```
skerschb: nit: since this logic of looking for "children" in the ast repeats a few times, perhaps a function that returns a boolean to wrap this.

snooty/semanticparser.py
```diff
@@ -54,3 +56,117 @@ def slug_title(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
...
+        # in find_toctree_nodes(...) is string rather than FileId.
+        for fileid in pages:
+            slug = fileid.without_known_suffix
+            fileid_dict[slug] = fileid
+
+        # Build the toctree
+        for fileid in pages:
+            child: Dict["str", Any] = {}
+            ast: Dict[str, Any] = cast(Dict[str, Any], pages[fileid].ast)
+            find_toctree_nodes(fileid, ast, pages, child, fileid_dict)
```
skerschb: nit: could look for children (In abstracted function) above find_toctree_nodes.

varsha804: I was able to remove the `for-loop` actually, so the function is much more simplified now!

snooty/semanticparser.py
```diff
@@ -8,21 +8,44 @@
...
+        self, fn_names: List[str]
+    ) -> List[Callable[[Dict[FileId, Page]], Dict[str, SerializableType]]]:
+        fn_list: List[Any] = []
+        for name in fn_names:
+            if name == "toctree":
+                fn_list.append(self.build_toctree)
+            elif name == "slug-title":
+                fn_list.append(self.build_slug_title)
+            elif name == "breadcrumbs":
+                fn_list.append(self.breadcrumbs)
```
sophstad: I think a string-to-function mapping may be a bit simpler than this conditional statement. Something like this:
```python
fn_mapping = {
    "toctree": self.build_toctree
    # etc.
}
```

snooty/semanticparser.py
```diff
@@ -79,20 +104,77 @@ def toctree(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
...
+            for i in range(len(reversed_path)):
+                slug = remove_leading_slash(path[i])
+                if path[:i]:
+                    if slug not in page_dict:
+                        page_dict[slug] = [path[:i]]
+                    else:
+                        if path[:i] not in page_dict[slug]:
+                            page_dict[slug].append(path[:i])
+
+        self.pages = {"pages": page_dict}
```
sophstad: Could we give this a more descriptive name? Perhaps `parentPaths`?

snooty/semanticparser.py
```diff
@@ -79,20 +104,77 @@ def toctree(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
...
+
+        return self.pages
+
+
+# Helper function used to retrieve the breadcrumbs for a particular slug
+def get_paths(root: Dict[str, Any], path: List[str], all_paths: List[Any]) -> None:
+    if not root:
+        return
+    if not children_exist(root) or children_exist(root) and len(root["children"]) == 0:
```
sophstad: [nit] Parentheses may clarify this a little
```suggestion
    if not children_exist(root) or (children_exist(root) and len(root["children"]) == 0):
```

snooty/semanticparser.py
```diff
@@ -144,16 +144,20 @@ def breadcrumbs(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
...
-                    if slug not in page_dict:
-                        page_dict[slug] = [path[:i]]
-                    else:
-                        if path[:i] not in page_dict[slug]:
-                            page_dict[slug].append(path[:i])
+
+                if slug not in page_dict:
+                    page_dict[slug] = [path[:i]]
+                else:
+                    if path[:i] not in page_dict[slug]:
```
sophstad: You could use a set here instead! Up to you~

sophstad: You could use a set here instead! Up to you~

snooty/semanticparser.py
varsha804: good idea!!

PR #57, opened 2019-11-08
DOCSP-7546: Internal Page Navigation (Semantic Parse)

varsha804:


snooty/test_semantic_parser.py
```diff
@@ -99,3 +99,9 @@ def test_breadcrumbs(backend: Backend) -> None:
 
     assert len(pages["page3"]) == 1
     assert pages["page3"] == ["page2"]
+
+
+def test_toctree_order(backend: Backend) -> None:
+    # Ensure that the correct pages and assets exist for toctree order
+    order: List[str] = cast(List[str], backend.metadata["toctreeOrder"])
+    assert order == ["/", "page1", "page2", "page3", "page1"]
```
sophstad: There shouldn't be any duplicates in this output! I also noticed dupes when running on `docs-ecosystem`. I suppose using a set instead of a list would fix, but perhaps there's an underlying problem in the recursion causing this.

varsha804: This outputs `page1` twice since it's a child of both `index` and `page3`, and therefore appears in the tree twice. That may be what's happening in `docs-ecosystem` as well. Do we still want to use a set in this case?

varsha804: Discussed offline: there shouldn't be any pages that have more than one parent, so we will change the file structure of the testing pages so that each non-root page has exactly one parent.

snooty/semanticparser.py
```diff
@@ -160,6 +161,25 @@ def breadcrumbs(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
 
         return self.parent_paths
 
+    def toctree_order(self, pages: Dict[FileId, Page]) -> Dict[str, SerializableType]:
```
sophstad: Definitely my bad for not requesting them all along, but I think a docstring would be useful here to clarify the purpose of this function.

varsha804: No worries! On it 😃

PR #60, opened 2019-11-11
DOCSP-6887: Update figure value to full local path when returning AST

rayangler:
[JIRA Ticket](https://jira.mongodb.org/browse/DOCSP-6887)

This PR changes the way figure nodes are handled before returning the page AST to Snooty on VS Code. The AST will now return figure nodes with values including their full path of the image on the user's local machine.

This PR must be merged before the VS Code PR for [DOCSP-6887](https://github.com/mongodb/snooty-vscode/pull/8).

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -797,13 +797,6 @@ class _Project:
         for page in pages:
             self._page_updated(page, diagnostic_list)
 
-        # Specify which transformations should be included in semantic postprocessing
-        fn_names: List[str] = ["toctree", "slug-title", "breadcrumbs", "toctree order"]
-        semantic_parse: Dict[str, SerializableType] = self.semantic_parser.run(
-            self.pages, fn_names
-        )
-        self.backend.on_update_metadata(self.prefix, semantic_parse)
-
     def delete(self, path: PurePath) -> None:
         file_id = os.path.basename(path)
         for giza_category in self.yaml_mapping.values():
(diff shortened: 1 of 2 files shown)


snooty/parser.py
```diff
@@ -797,13 +797,6 @@ def create_page() -> Tuple[Page, EmbeddedRstParser]:
         for page in pages:
             self._page_updated(page, diagnostic_list)
 
-        # Specify which transformations should be included in semantic postprocessing
-        fn_names: List[str] = ["toctree", "slug-title", "breadcrumbs"]
-        semantic_parse: Dict[str, SerializableType] = self.semantic_parser.run(
-            self.pages, fn_names
-        )
-        self.backend.on_update_metadata(self.prefix, semantic_parse)
```
rayangler: Deleted due to performance issues on Snooty VS Code. Changing which file the editor is focused on ends up causing the extension to lag and be unusable.

snooty/parser.py
```diff
@@ -893,15 +886,35 @@ def replace_include(node: Dict[str, Any]) -> Dict[str, Any]:
...
 
                 # Resolve includes within include node
-                replaced_include = list(map(replace_include, include_node_children))
+                replaced_include = list(map(replace_nodes, include_node_children))
                 node["children"] = replaced_include
+            # Replace instances of an image's name with its full path. This allows Snooty Preview to render an image by
+            # using the location of the image on the user's local machine
+            elif node.get("name") == "figure":
+                # Obtain subset of the image's path (name)
+                argument = node["argument"][0]
```
skerschb: I'm assuming that elsewhere in the parser we've made certain that we don't ever get a figure without a first argument? Just checking.

These PR looks really fine to me -- but for some reason when I tried to run the tests locally, I got a failure.

rayangler: Yup, this is checked in [`parser.py`](https://github.com/mongodb/snooty-parser/pull/60/files#diff-84e66ec308ed95ea541bded28cfbdff1L410-L416)!
```
if argument_text is None:
                self.diagnostics.append(
                    Diagnostic.error(
                        f'"{name}" expected a path argument', util.get_line(node)
                    )
                )
                return True
```

PR #66, opened 2019-11-20
DOCSP-7497: Save formatting in slug-title mapping

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-7497)]
- Include array of text nodes in slug-title mapping to easily render formatted text in the TOC sidebar and breadcrumbs
- Replace current implementation with implementation using event parser
- Also adds a handler for page entry to the event parser, which I forgot to add in #64


snooty/semanticparser.py
```diff
@@ -37,67 +43,42 @@ def run_event_parser(
...
-            slug_title_dict[slug] = title
-
-        self.slug_title = {"slugToTitle": slug_title_dict}
-
-        return self.slug_title
+    def build_slug_title_mapping(
+        self, filename: FileId, *args: SerializableType, **kwargs: SerializableType
+    ) -> None:
+        """Construct a slug-title mapping of all pages in property"""
+        obj = cast(Dict[str, SerializableType], kwargs.get("obj"))
```
i80and: `kwargs.get("obj")` can return None; I would rather that it be casted to `Optional[Dict[str, SerializableType]]`, then have `assert obj is not None` to refine the type

snooty/semanticparser.py
```diff
@@ -37,67 +43,42 @@ def run_event_parser(
...
+        slug = filename.without_known_suffix
+
+        # Don't parse these files for headings
+        if "includes" in slug or "images" in slug:
```
i80and: Don't we just want to scan `.txt` files, allowing us to ask `if filename.suffix == ".txt"`?

If this is really what we want, it may still be more broad than we'd like: perhaps `if "includes" in filename.parts or "images" in filename.parts`?

snooty/semanticparser.py
```diff
@@ -37,67 +43,42 @@ def run_event_parser(
...
+            return
+
+        # Save the first heading we encounter to the slug title mapping
+        if self.slug_title_mapping.get(slug) is None and obj.get("type") == "heading":
```
i80and: `if slug not in self.slug_title_mapping and...` would be more idiomatic

snooty/semanticparser.py
```diff
@@ -37,67 +43,42 @@ def run_event_parser(
...
+            self.slug_title_mapping[slug] = cast(
```
i80and: See previous remark on `obj.get` returning an Optional we should assert away

snooty/semanticparser.py
sophstad: Yes you're right! The first bit is exactly what we want, thanks 👍

snooty/semanticparser.py
```diff
@@ -227,14 +208,14 @@ def find_toctree_nodes(
     pages: Dict[FileId, Page],
     node: Dict[Any, Any],
     fileid_dict: Dict[str, FileId],
-    slug_title: Dict[str, Any],
+    slug_title_mapping: Dict[str, Any],
```
i80and: Should this be `Dict[str, List[SerializableType]]`?

PR #67, opened 2019-11-21
DOCSP-7800: Prepare for automated releases

i80and:


i80and:
Packages are currently built for EVERY push, which may not be desirable. Thoughts?

i80and:
@sophstad ping

PR #68, opened 2019-11-22
DOCSP-7187: Parser returns AST of .rst file in a pseudo page AST object

rayangler:
[JIRA Ticket](https://jira.mongodb.org/browse/DOCSP-7187)

* Parser returns AST of a .rst file as if it was included as a child of an empty AST object.


snooty/language_server.py
```diff
@@ -300,14 +300,47 @@ def m_text_document__resolve(
...
+        file containing a single include directive to said .rst file.
+        """
+
         if self.project is None:
             logger.warn("Project uninitialized")
             return None
 
-        return self.project.get_page_ast(Path(filePath))
+        filePath = Path(fileName)
+        page_ast = self.project.get_page_ast(filePath)
```
i80and: jooc what happens if this fails?

PR #70, opened 2019-12-04
Bugfix: Don't overwrite toctree titles

sophstad:
Fixes bug where user-specified toctree titles were being overwritten. Now, check to see if a title has been set. If not, attempt to lookup the page title to use. Otherwise, assign title to `None`.

Bug identified in https://github.com/mongodb/docs-node/pull/4.

### DIFF
--- a/snooty/semanticparser.py
+++ b/snooty/semanticparser.py
@@ -216,7 +216,7 @@ def find_toctree_nodes(
         if node:
             node["slug"] = fileid.without_known_suffix
             # Look up title in slug-title mapping if it has not been specified
-            if "title" not in node or node.get("title") is None:
+            if node.get("title") is None:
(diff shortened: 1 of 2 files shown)


snooty/semanticparser.py
```diff
@@ -215,7 +215,9 @@ def find_toctree_nodes(
     if not children_exist(ast):
         if node:
             node["slug"] = fileid.without_known_suffix
-            node["title"] = slug_title_mapping.get(node["slug"], [])
+            # Look up title in slug-title mapping if it has not been specified
+            if "title" not in node or node.get("title") is None:
```
i80and: Is this redundant? `node.get("title")` will return None if `"title" not in node`

sophstad: Yes! Thanks!

PR #73, opened 2019-12-11
Refactor semantic parse layer

sophstad:
I started refactoring `find_toctree_nodes()` based on [this comment](https://github.com/mongodb/snooty-parser/pull/72#discussion_r356050444), but then realized there was a lot of room for improvement 😅

### DIFF
--- a/snooty/semanticparser.py
+++ b/snooty/semanticparser.py
@@ -12,6 +12,11 @@ ELEMENT_EVENT = "element"
 
 
 class SemanticParser:
+    """SemanticParser handles all operations on parsed AST files.
+
+    The only method that should be called on an instance of SemanticParser is run(). This method
+    handles calling all other methods and ensures that parse operations are run in the correct order."""
+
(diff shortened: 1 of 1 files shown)


snooty/semanticparser.py
```diff
@@ -85,201 +77,167 @@ def build_slug_title_mapping(
...
-        # Construct a {slug: fileid} mapping so that we can retrieve the full file name
-        # given a slug. We cannot use the with_suffix method since the type of the slug
-        # in find_toctree_nodes(...) is string rather than FileId.
-        for fileid in pages:
-            slug = fileid.without_known_suffix
-            fileid_dict[slug] = fileid
-
         if not self.slug_title_mapping:
-            self.run_event_parser(pages)
+            self.run_event_parser()
```
i80and: I see a couple of these `if not self.foo: <generate-self.foo>` constructs. Is this class intended to be used in a one-shot sort of way? If so, we probably want to mention that in the class docstring.

PR #76, opened 2019-12-18
points node-api and node-docs roles to the 3.3 docs

schmalliso:
We're documenting some methods that don't exist in the 3.0 driver, so we'd like to point the node-api/node-docs roles to 3.3. Tweaking it myself seems like it would create less work for y'all.


i80and:
@schmalliso looks like there's a merge conflict! If you rebase, I can go ahead and merge :+1: Thanks for doing this!

PR #78, opened 2019-12-18
DOCSP-8084: Improve yaml output_file handling

i80and:
* `Page.output_filename` becomes non-Optional, and a `Page.create()` class method is added.
* All output filename hacking is removed.
* `GizaCategory.to_pages()` takes an additional source path parameter, and the page creation factory takes in an output filename, allowing each GizaCategory to specify its desired output filename.

### DIFF
--- a/snooty/gizaparser/steps.py
+++ b/snooty/gizaparser/steps.py
@@ -126,10 +126,13 @@ class GizaStepsCategory(GizaCategory[Step]):
 
     def to_pages(
         self,
-        page_factory: Callable[[], Tuple[Page, EmbeddedRstParser]],
+        source_path: Path,
+        page_factory: Callable[[str], Tuple[Page, EmbeddedRstParser]],
         steps: Sequence[Step],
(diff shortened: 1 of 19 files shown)


snooty/gizaparser/steps.py
```diff
@@ -126,10 +126,13 @@ def parse(
...
     def to_pages(
         self,
-        page_factory: Callable[[], Tuple[Page, EmbeddedRstParser]],
+        source_path: Path,
+        page_factory: Callable[[str], Tuple[Page, EmbeddedRstParser]],
         steps: Sequence[Step],
     ) -> List[Page]:
-        page, rst_parser = page_factory()
+        output_filename = source_path.with_suffix(".rst").name
+        output_filename = output_filename[6:]
```
sophstad: What does `6` refer to here? Getting magic number vibes~

i80and: Made this more clear. Thanks!

PR #81, opened 2020-01-08
Decouple primary key from page ID

sophstad:
Dependency for [DOCSP-8277](https://jira.mongodb.org/browse/DOCSP-8277).
- Use MongoDB's standard `ObjectID` as primary key in Snooty's `documents` and `metadata` collections
- Write to database based on identified environment
  - Environment variable `SNOOTY_ENV` must be set to one of these options:
    1. `development`
    2. `staging`
    3. `production`
  - Environment can be configured on autobuilder using Docker environment variables (e.g. `--env SNOOTY_ENV=staging`)
  - Environment defaults to `development`

This will futureproof the data and allow us to create unique compound indexes in order to enforce build uniqueness, rather than rely on the specific composition of the `_id` field.

Also includes small improvements to modularize collection names.


snooty/main.py
```diff
@@ -89,6 +100,21 @@ class MongoBackend(Backend):
...
+            db_name = config["environments"][SNOOTY_ENV]["db"]
+            assert isinstance(db_name, str)
+            return db_name
+
+    def _manage_indexes(self) -> None:
+        # List of indexes to be created. For now, metadata and documents collections use the same indexes.
+        indexes = [IndexModel("page_id", unique=True)]
+
+        self.client[self.db][COLL_DOCUMENTS].create_indexes(indexes)
+        self.client[self.db][COLL_METADATA].create_indexes(indexes)
```
skerschb: I'm not clear on the benefits of creating an index here? I'm imagining we will set this at db setup.

sophstad: Yes true, this can be removed!

stubs/pymongo/collection.pyi
```diff
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
```
i80and: ?

PR #84, opened 2020-01-10
DOCSP-6657: Add local targets to the TargetDatabase

i80and:
There's one big strawman AST structure change: `target` and `target_directive` are unified. So:
`<target ids=[...] uriref="...">`
becomes
`<target domain="..." name="..." target="..." uriref="...">`
and `target_directive` becomes regular `target`.
Let me know what you think of this idea.

There's also a caveat in that the following things won't work right:
* `option` (requires special parsing voodoo)

### DIFF
--- a/snooty/gizaparser/test_steps.py
+++ b/snooty/gizaparser/test_steps.py
@@ -56,7 +56,7 @@ def test_step() -> None:
             "following command to import the\n</text><reference ",
             'refuri="https://www.mongodb.org/static/pgp/server-3.4.asc">',
             "<text>MongoDB public GPG Key</text>",
-            "</reference><target ids=\"['mongodb-public-gpg-key']\" ",
+            """</reference><target domain="std" name="label" target="mongodb-public-gpg-key" """,
             'refuri="https://www.mongodb.org/static/pgp/server-3.4.asc">',
             "</target></paragraph></section></directive>",
             '<directive name="step">',
--- a/snooty/semanticparser.py
+++ b/snooty/semanticparser.py
@@ -1,6 +1,6 @@
 import re
 from collections import defaultdict
-from typing import Any, Callable, cast, Dict, List, Optional, Set, Tuple
+from typing import Any, Callable, cast, Dict, List, Optional, Set, Tuple, Iterable
 from .types import (
     Diagnostic,
     FileId,
@@ -48,8 +48,18 @@ class SemanticParser:
 
         document: Dict[str, SerializableType] = {}
 
+        self.run_event_parser(
+            [
+                (OBJECT_START_EVENT, self.populate_include_nodes),
+                (OBJECT_START_EVENT, self.build_slug_title_mapping),
+                (OBJECT_START_EVENT, self.handle_target),
+            ]
+        )
+
         # Update metadata document with key-value pairs defined in event parser
-        document.update(self.run_event_parser())
+        document.update({"slugToTitle": self.slug_title_mapping})
+
+        self.run_event_parser([(OBJECT_START_EVENT, self.validate_ref_targets)])
 
         # Run semantic parse operations related to toctree and append to metadata document
         document.update(
@@ -62,17 +72,14 @@ class SemanticParser:
 
         return document, self.diagnostics
 
-    def run_event_parser(self) -> Dict[str, SerializableType]:
+    def run_event_parser(
+        self, listeners: Iterable[Tuple[str, Callable[[Any], None]]]
+    ) -> None:
         event_parser = EventParser()
-        event_parser.add_event_listener(OBJECT_START_EVENT, self.populate_include_nodes)
-        event_parser.add_event_listener(
-            OBJECT_START_EVENT, self.build_slug_title_mapping
-        )
-        event_parser.add_event_listener(OBJECT_START_EVENT, self.validate_ref_targets)
-        event_parser.consume(self.pages)
+        for event, listener in listeners:
+            event_parser.add_event_listener(event, listener)
 
-        # Return dict containing fields updated in event-based parse
-        return {"slugToTitle": self.slug_title_mapping}
+        event_parser.consume(self.pages.items())
 
     def validate_ref_targets(
         self,
@@ -92,9 +99,11 @@ class SemanticParser:
             name = obj.get("name")
             target = obj.get("target")
             key = f"{domain}:{name}:{target}"
+
             try:
                 # Append URL to AST
-                obj["url"] = self.targets.get_url(key)
+                field_name, target = self.targets.get_url(key)
+                obj[field_name] = target
             except KeyError:
                 position = cast(Any, obj.get("position"))
                 start = position["start"]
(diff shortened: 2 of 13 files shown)


snooty/test_language_server.py
```diff
@@ -134,7 +134,7 @@ def test_text_doc_get_page_ast() -> None:
         # Change image path to be full path
         index_ast_string = (
             """<root>
-            <target ids="['guides']"></target>
+            <target domain="std" name="label" target=\"guides\"></target>
```
sophstad: Why `target=\"guides\"` output with the extra backslashes? The other fields don't include them.

snooty/types.py
```diff
@@ -178,7 +178,7 @@ class TargetDatabase:
     """A database of targets known to this project."""
 
     intersphinx_inventories: Dict[str, intersphinx.Inventory]
-    local_definitions: Dict[str, str]
+    local_definitions: Dict[str, Tuple[FileId, str]]
```
sophstad: What is the purpose of the `str` in the tuple? It looks like it is only ever set to an empty string, and never accessed.

snooty/types.py
```diff
@@ -190,17 +190,27 @@ def __contains__(self, key: str) -> bool:
 
         return False
 
-    def get_url(self, key: str) -> str:
-        # TODO: add check for local definitions here
+    def get_url(self, key: str) -> Tuple[str, str]:
+        # Check to see if the target is defined locally
+        try:
+            return ("fileid", self.local_definitions[key][0].as_posix())
```
sophstad: Wondering if there is any benefit to including the file suffix in the target node? If not, we could use FileId's [`without_known_suffix()`](https://github.com/mongodb/snooty-parser/blob/master/snooty/types.py#L47-L51) method.

snooty/test_semantic_parser.py
```diff
@@ -94,18 +97,40 @@ def test_expand_includes(backend: Backend) -> None:
...
+    # Assert that targets found in intersphinx inventories are correctly resolved
     paragraph = cast(Dict[str, Any], ast["children"][0])
     ref_role = paragraph["children"][0]
-    assert ref_role["type"] == "ref_role"
-    assert "url" in ref_role
-    assert (
-        ref_role["url"]
-        == "https://docs.mongodb.com/manual/reference/configuration-options/#net.port"
+    print(ast_to_testing_string(ref_role))
+    check_ast_testing_string(
```
sophstad: Thanks for this improvement 🙌 I think there are other places where it could be used in this file as well. Alas!

snooty/test_parser.py
```diff
@@ -571,12 +571,13 @@ def test_rstobject() -> None:
     )
     page.finish(diagnostics)
     assert diagnostics == []
+    print(ast_to_testing_string(page.ast))
```
sophstad: Delete?

snooty/test_semantic_parser.py
```diff
@@ -94,18 +97,40 @@ def test_expand_includes(backend: Backend) -> None:
...
+
+    # Assert that targets found in intersphinx inventories are correctly resolved
     paragraph = cast(Dict[str, Any], ast["children"][0])
     ref_role = paragraph["children"][0]
-    assert ref_role["type"] == "ref_role"
-    assert "url" in ref_role
-    assert (
-        ref_role["url"]
-        == "https://docs.mongodb.com/manual/reference/configuration-options/#net.port"
+    print(ast_to_testing_string(ref_role))
```
sophstad: Delete?

snooty/test_semantic_parser.py
```diff
@@ -94,18 +97,40 @@ def test_expand_includes(backend: Backend) -> None:
...
+        domain="mongodb"
+        name="setting"
+        target="net.port"
+        url="https://docs.mongodb.com/manual/reference/configuration-options/#net.port"></ref_role>""",
+    )
+
+    # Assert that local targets are correctly resolved
+    paragraph = cast(Dict[str, Any], ast["children"][3])
+    ref_role = paragraph["children"][1]
+    print(ast_to_testing_string(ref_role))
```
sophstad: Delete?

snooty/gizaparser/test_steps.py
```diff
@@ -56,7 +56,7 @@ def create_page(filename: Optional[str]) -> Tuple[Page, EmbeddedRstParser]:
             "following command to import the\n</text><reference ",
             'refuri="https://www.mongodb.org/static/pgp/server-3.4.asc">',
             "<text>MongoDB public GPG Key</text>",
-            "</reference><target ids=\"['mongodb-public-gpg-key']\" ",
+            """</reference><target domain="std" name="label" target="mongodb-public-gpg-key" """,
```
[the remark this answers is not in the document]
i80and: Yup! Triple quotes avoid the need to escape the double-quotes inside

snooty/test_language_server.py
i80and: Oh, good catch. I was escaping them, which is unnecessary because it's a triple-quote string!

snooty/types.py
i80and: I don't see any need for the file suffix; happy to remove it!

snooty/test_semantic_parser.py
```diff
@@ -44,7 +45,9 @@ def test_expand_includes(backend: Backend) -> None:
         {
             "type": "target",
             "position": {"start": {"line": 1}},
-            "ids": ["connection-limits"],
+            "domain": "std",
+            "name": "label",
```
sophstad: It looks like the `name` field is always being output as `label` (in these tests and after running the parser on my machine). Is this expected?

i80and: That'll be true for any targets defined using the `.. _foo:` syntax

skerschb:
Of the things mentioned:

Glossary terms (requires special parsing voodoo)
option (requires special parsing voodoo)
Any rstobject using Giza's wacky prefix thing (e.g. binary)

how many of these are requirements for ecosystem/docs-node

snooty/types.py
i80and: ugh forgot to fill that in. The str is meant to point to the ID of the element to link to

i80and: @sophstad Does the frontend have enough information to generate a complete link based on the page name and target?

sophstad: @i80and Based on our discussion last week, I thought more would be implemented in the parser layer. Namely, I thought that the parser would be responsible for fetching the ID's title (the link text) and including it in the node. I do think that is a little bit much for the front end to take on, but am open to other approaches!

i80and: Fetching the title will be done in the formatting ticket, I think! My concern is about the href right now

sophstad: Oh gotcha! Yes, this looks adequate for the frontend.

snooty/semanticparser.py
```diff
@@ -48,8 +48,18 @@ def run(
...
+                (OBJECT_START_EVENT, self.build_slug_title_mapping),
+                (OBJECT_START_EVENT, self.handle_target),
+            ]
+        )
+
         # Update metadata document with key-value pairs defined in event parser
-        document.update(self.run_event_parser())
+        document.update({"slugToTitle": self.slug_title_mapping})
+
+        self.run_event_parser([(OBJECT_START_EVENT, self.validate_ref_targets)])
```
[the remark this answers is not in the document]
i80and: It does, but the error is duplicated! Interesting!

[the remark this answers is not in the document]
i80and: Sue says that this doesn't need to block this PR :+1:

PR #90, opened 2020-01-27
DOCSP-7988: Add support for project title

mango-db:


snooty/semanticparser.py
```diff
@@ -226,7 +226,8 @@ def build_toctree(self) -> Dict[str, SerializableType]:
 
         # Build the toctree
         root: Dict[str, SerializableType] = {
-            "title": self.project_config.name,
+            "title": self.project_config.title,
+            "name": self.project_config.name,
```
sophstad: We actually don't need `name` in order to render the toctree, so you can remove this line. But good thinking!

snooty/types.py
```diff
@@ -401,7 +402,7 @@ def open(cls, root: Path) -> Tuple["ProjectConfig", List[Diagnostic]]:
 
             path = path.parent
 
-        return cls(root, "untitled"), diagnostics
+        return cls(root, name="unnamed", title="untitled"), diagnostics
```
i80and: Because `title` has a default value, you don't actually need to pass that parameter

snooty/types.py
```diff
@@ -365,6 +365,7 @@ def finish(
 class ProjectConfig:
     root: Path
     name: str
+    title: str = field(default="untitled")
```
sophstad: In this case, I think we actually _don't_ want to provide a default value. I believe that `title` should be a required field in the same way that `name` is, because I can't imagine a scenario where a site doesn't need a title. Thoughts @i80and?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support for defining a project's verbose title (DOCSP-7988)
```
i80and: I'm not sure `verbose` adds any meaning

snooty/semanticparser.py
```diff
@@ -256,7 +257,7 @@ def find_toctree_nodes(
                 toctree_node: Dict[str, SerializableType] = {}
                 if "url" in entry:
                     toctree_node = {
-                        "title": entry.get("title", None),
+                        "title": entry.get("title"),
```
sophstad: This section of code relates to toctree node titles, not the site title, so it is unrelated to your work. This change shouldn't be included in the PR!

PR #91, opened 2020-01-31
DOCSP-8667: Add performance reporting system

i80and:


.gitignore
```diff
@@ -9,3 +9,4 @@ node_modules/
 htmlcov/
 .venv/
 dist/
+test_data/docs/
```
sophstad: Add `.docs/` to `.gitignore`

i80and: oop, yes, thank you!

PR #92, opened 2020-01-31
DOCSP-7399: Temporarily remove multiprocessing

i80and:
We will need to unify pending tasks and postprocessing to work with
Python 3.8's macOS new default multiprocessing spawn mode.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -35,6 +35,10 @@ from .types import (
     BuildIdentifierSet,
 )
 
+# XXX: Work around to get snooty working with Python 3.8 until we can fix
+# our implicit data flow issues.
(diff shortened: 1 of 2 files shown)


snooty/parser.py
```diff
@@ -35,6 +35,10 @@
     BuildIdentifierSet,
 )
 
+# XXX: Work around to get snooty working with Python 3.8 until we can fix
```
sophstad: Is there a convention around `XXX`? This seems kind of like a TODO.

i80and: XXX indicates something that's gross or a bug; basically the same as FIXME

https://softwareengineering.stackexchange.com/questions/65467/what-does-xxx-mean-in-a-comment

PR #94, opened 2020-02-04
DOCSP-8723: Add devhub directives

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-8723)]
- Adds support for `default_domain` field in snooty.toml
- Adds the following directives to the `devhub` domain:
  - `meta-description`
  - `series`
  - `author`
  - `aft-image`
  - `type`
  - `level`
  - `tags`
  - `products`
  - `introduction`
  - `prerequisites`
  - `content`
  - `summary`
  - `related`
- Adds the following directives to the empty domain:
  - `pubdate`
  - `twitter`
  - `youtube`
  - `twitch`


snooty/rstparser.py
```diff
@@ -696,10 +696,17 @@ class Domain:
...
 class Registry:
+    default_domain: Optional[str] = None
     domains: DefaultDict[str, Domain] = field(
         default_factory=lambda: defaultdict(Domain)
     )
 
+    def get_domain_resolution_sequence(self) -> List[str]:
+        # If default domain has been specified, prepend it to the list of domains
+        if self.default_domain:
+            return [self.default_domain] + DOMAIN_RESOLUTION_SEQUENCE
```
i80and: This will involve constructing *two* new list instances upon every directive/role lookup. This should probably be a precomputed attribute of the Registry.

Bonus points if directive/role resolution is factored into methods of this class, maybe?

snooty/rstspec.toml
```diff
@@ -8,6 +8,8 @@ guide_categories = ["Getting Started"]
 alignment = ["left", "center", "right"]
 card_type = ["small", "large"]
 output_format = ["html"]
+devhub_type = ["article", "quickstart", "how-to", "video", "live"]
+difficulty = ["beginner", "intermediate", "advanced"]
```
i80and: If this is intended to be a distinct enum from `user_level`, could we prefix with `devhub_`?

i80and:
Should `author-image` be an option of `author`? cc @schmalliso

snooty/rstspec.toml
sophstad: Oh duh. I'm happy to use the existing directive and delete `devhub:level`.

snooty/rstparser.py
sophstad: Agree with this! The only reason I didn't do this is because I had some trouble figuring out how to do that within the confines of `@dataclass`. Maybe the most effective way to fix this would be for you to pull this branch down and make that edit, given that we are out tomorrow. Let me know!

i80and: While it is possible to have a dataclass with a constructor, this doesn't actually need to be a dataclass. Remove the `@dataclass` decorator, and just have a normal constructor instead of the field declarations.

eshwer:
Is `Date` something we are going to automate or have users input? 
Also, would there be a specific directive for `Author Bio`

schmalliso:
I would happy for author-image to be an option of author!

skerschb:
Just a note about these author modifier fields -- we actually don't want to encourage authors to put their data into each article. I'm assuming we are capturing this here so that we make it possible to create an author document within the devrel content. But that we would perhaps include that document in the render template based on an author doc lookup. Are we all in agreement?

sophstad:
> Just a note about these author modifier fields -- we actually don't want to encourage authors to put their data into each article. I'm assuming we are capturing this here so that we make it possible to create an author document within the devrel content. But that we would perhaps include that document in the render template based on an author doc lookup. Are we all in agreement?

@skerschb That makes sense. So for example, articles could include a document called `includes/authors/author-name.rst` that includes something like this:
```
.. author: Author Name
   :image: path/to/image.jpg
   :bio: Short bio here
```

@eshwer AFAIK they don't want date to be automatically generated based on when the doc is created. They will write the date here and we can massage it into an accurate date time format in the parser level. And see above, I think we can include `bio` as an option of `author`.

snooty/rstparser.py
sophstad: I tried getting into that briefly yesterday and ran in to trouble. I will leave as-is for now and mark with an `XXX` comment!

PR #97, opened 2020-02-06
DOCSP-8731: Support page groups

i80and:
This adds an optional `pageGroups` field to the build metadata containing cleaned and validated page group slugs.

Example `snooty.toml` syntax:
```toml
[page_groups]
"Group 1" = ["index", "foo", "/page4"]
```

Example output field in the metadata document:
```json
	"pageGroups" : {
		"A Test Group" : [
			"index",
			"platforms",
			"tutorial/getting-started-with-hadoop"
		]
	}
```

The careful reader will note that issues are logged rather than sent to the normal diagnostic mechanism: this is because our diagnostic mechanism doesn't currently handle snooty.toml issues properly (currently the vscode extension crashes), and I didn't want to make that problem any worse.

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -1,4 +1,5 @@
 import re
+import logging
 from collections import defaultdict
 from typing import Any, Callable, cast, Dict, List, Optional, Tuple, Iterable, Sequence
 from .eventparser import EventParser
@@ -12,7 +13,7 @@ from .types import (
 )
 from .util import get_child_of_type
 
-
+logger = logging.getLogger(__name__)
 PAT_FILE_EXTENSIONS = re.compile(r"\.((txt)|(rst)|(yaml))$")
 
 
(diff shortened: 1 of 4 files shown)


snooty/postprocess.py
```diff
@@ -104,6 +105,30 @@ def run(
...
+                slug
+                for slug in (clean_and_validate_page_group_slug(slug) for slug in slugs)
+                if slug
+            ]
+            for title, slugs in self.project_config.page_groups.items()
+        }
+
+        if page_groups:
+            document.update({"pageGroups": page_groups})
+
```
sophstad: Because I don't know of any plans to use this type of content in Docs, it may make sense instead include this in `DevhubPostprocessor.run` from #96 and avoid including it in a normal docs parse. Thoughts?

snooty/test_postprocess.py
```diff
@@ -242,3 +242,9 @@ def test_toctree_order(backend: Backend) -> None:
     # Ensure that the correct pages and assets exist for toctree order
     order: List[str] = cast(List[str], backend.metadata["toctreeOrder"])
     assert order == ["/", "page1", "page2", "page3"]
+
+
+def test_page_groups(backend: Backend) -> None:
+    """Test that page groups are correctly filtered and cleaned."""
+    page_groups: Dict[str, List[str]] = cast(Any, backend.metadata["pageGroups"])
+    assert page_groups == {"Group 1": ["index", "page4"]}
```
sophstad: Ditto with adding this test to a new `test_devhub.py` file.

test_data/test_postprocessor/snooty.toml
```diff
@@ -5,3 +5,6 @@ intersphinx = ["https://docs.mongodb.com/manual/objects.inv"]
 [constants]
 name = "MongoDB name"
 title = "MongoDB title"
+
+[page_groups]
+"Group 1" = ["index", "foo", "/page4"]
```
sophstad: And move this test data as well!

snooty/postprocess.py
i80and: Makes sense! I'll rebase as soon as #96 is merged!

PR #99, opened 2020-02-06
DOCSP-8799: Fix tests & refactor where needed

i80and:


snooty/test_legacy_guides.py
```diff
@@ -68,3 +68,4 @@ def test_legacy_guides() -> None:
     # Ensure that legacy syntax and new syntax create the same output
     check_ast_testing_string(legacy_page.ast, correct)
     check_ast_testing_string(new_page.ast, correct)
+    # assert False
```
sophstad: Just wondering what this is!

i80and: ah drat I was using that to make sure pytest printed things, but forgot to remove it. Thank you!

PR #108, opened 2020-02-24
DOCSPLAT-114: Handle substitutions in postprocess layer

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSPLAT-114)] Substitutions were previously handled by a combination of `parser.py` and the front-end. Now, handle almost everything in the postprocess layer.


snooty/parser.py
```diff
@@ -701,7 +687,10 @@ def __init__(
         substitution_nodes: Dict[str, SerializableType] = {}
         for k, v in self.config.substitutions.items():
             node, substitution_diagnostics = parse_rst(self.parser, root, v)
-            substitution_nodes[k] = node.ast
+            ast = cast(Any, node.ast)
+            # Extract text nodes from <root> and <paragraph> parent nodes
+            paragraph = ast["children"][0]
```
i80and: What happens if an empty substitution is defined?

test_data/test_postprocessor/source/page4.txt
```diff
@@ -1 +1,7 @@
 .. include:: /includes/test.rst
+
+Using |sub| before definition.
+
+.. |sub| replace:: Substitution
+
+Using |sub| after definition.
```
i80and: I think we optionally might consider making this test a little more comprehensive:
* Define a substitution in an include, use it here
* Use `|sub|` in an include
* If `|sub|` is redefined elsewhere in the page, what evaluation order is used in determining the value at any given point?

sophstad: Good points here! Regarding point 3, current behavior is that the substitution definition is overwritten whenever the same `|sub|` is encountered. Overwriting is based on the parse order of the page which is... I'm not sure! Ultimately all references to `|sub|` are replaced with the same value. Do you think this is logical behavior? For the record, implementing another way seems potentially _much_ harder.

snooty/parser.py
sophstad: Added handling!

PR #109, opened 2020-02-25
Remove series directive

sophstad:


schmalliso:
@i80and We do! Thanks for the flag.

PR #116, opened 2020-03-03
DOCSPLAT-861: Fix image handling for devhub

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSPLAT-861)]
- Fix handling for author and ATF images
- Extract data from directives more carefully in postprocess layer


snooty/parser.py
```diff
@@ -517,6 +518,26 @@ def handle_directive(
                 self.diagnostics.append(Diagnostic.error(msg, util.get_line(node)))
         elif name == "toctree":
             doc["entries"] = node["entries"]
+        elif domain == "devhub" and name == "author":
+            image_argument = options.get("image", None)
```
i80and: The second argument of `dict.get()` actually defaults to None, so you don't need to include that

snooty/parser.py
```diff
@@ -517,6 +518,26 @@ def handle_directive(
...
+
+            if image_argument is None:
```
i80and: I might actually use `if not image_argument:`

PR #121, opened 2020-03-14
Handle <problematic> nodes

i80and:
After the AST refactor, building ecosystem fails because there are unknown roles causing docutils to emit "problematic" nodes, triggering an assertion in the parser.

This PR adds a test and suppresses the node, but I'm open to other behaviors!


sophstad:
Why didn’t Actions run on this PR?

i80and:
A dinnae ken! GitHub just says "One or more errors occurred" without the usual option to restart the check

PR #132, opened 2020-04-01
Dop 878: Refactor Snooty Diagnostic Usage

madelinezec:
Relevant docs: 

[Current issues for which snooty-parser raises diagnostics](https://docs.google.com/document/d/1775tlj0-EkL66nIbEAHECQ1kKiDGSZzjEAiNFTvpif0/edit?usp=sharing)
[Diagnostic Class Design](https://docs.google.com/document/d/1lkU1bEWAn1A4Jqo7iGPdoQSh5kiz8lmv6hAl_7VOM90/edit?usp=sharing)


snooty/types.py
```diff
@@ -82,65 +105,141 @@ def from_docutils(cls, docutils_level: int) -> "Diagnostic.Level":
             level = max(level, cls.info)
             return cls(level)
 
-    severity: Level
-    message: str
-    start: Tuple[int, int]
-    end: Tuple[int, int]
+    @property
+    # why is the below incorrect?
+    # def severity(self) -> Level:
```
i80and, line 110: It needs to be `-> "Diagnostic.Level":`, with quotes: this is a side-effect of bolting static types onto a language over 25 years after the language was born. The name `Diagnostic` hasn't actually been created yet, and class attributes cannot access other class attributes at time of instantiation

snooty/types.py
```diff
@@ -66,10 +66,33 @@ def without_known_suffix(self) -> str:
...
+
+        if end is None:
+            end_line, end_column = start_line, 1000
+        elif isinstance(end, int):
+            end_line, end_column = end, 1000
+        else:
+            end_line, end_column = end
+        self.end = (end_line, end_column)
+
+    # class attribute
```
i80and, line 95: This kind of comment isn't really necessary

snooty/types.py
```diff
@@ -66,10 +66,33 @@ def without_known_suffix(self) -> str:
         return fileid.as_posix()
 
 
-@dataclass
 class Diagnostic:
-    __slots__ = ("message", "severity", "start", "end")
+    # instance properties
+    __slots__ = ("message", "start", "end")
```
i80and, line 71: Just drop the __slots__ attribute, and the preceding comment. It's a space optimization that isn't actually necessary here, and won't really work anymore anyway with subclasses unless you add __slots__ to all of them as well

snooty/postprocess.py
```diff
@@ -277,7 +281,8 @@ def replace_substitutions(self, filename: FileId, node: n.Node) -> None:
                     del self.substitution_definitions[node.name]
                     node.children = []
                     self.diagnostics[filename].append(
-                        Diagnostic.error(
+                        # dont actually know if this is ok - but I think so?
```
i80and, line 284: Yup, this is fine

snooty/types.py
```diff
@@ -66,10 +66,33 @@ def without_known_suffix(self) -> str:
         return fileid.as_posix()
 
 
-@dataclass
 class Diagnostic:
-    __slots__ = ("message", "severity", "start", "end")
+    # instance properties
```
i80and, line 70: This is a bit of a refactor, but I actually think Diagnostic & all of its children should be moved to an `diagnostics.py` file, if you're up for that.

snooty/types.py
```diff
@@ -82,65 +105,141 @@ def from_docutils(cls, docutils_level: int) -> "Diagnostic.Level":
...
+
+class ErrorLoadingFile(LoadDiagnostic):
+    severity = Diagnostic.Level.error
+
+
+class OSDiagnostic(LoadDiagnostic):
+    pass
+
+
+class CannotOpenFile(OSDiagnostic):
```
i80and, line 241: Let's actually mock out making this a little more interesting!

```
class CannotOpenFile(OSDiagnostic):
    def __init__(self, path: Path, reason: str, start: ..., end: ...) -> None:
        super().__init__(self, "Error opening {str(path)}: {reason}", start, end)
        self.path = path
        self.reason = reason
```

Then, you don't need to construct the reason each time it's instantiated: you just construct e.g. `CannotOpenFile(parent_identifier.file, os.strerror(errno.ENOENT), start)` or `CannotOpenFile(self.asset.fileid, err.strerror, start)`.

How does this sound to you?

snooty/test_parser.py
```diff
@@ -200,7 +203,7 @@ def test_literalinclude() -> None:
 """,
     )
     page.finish(diagnostics)
-    assert len(diagnostics) == 1
+    assert len(diagnostics) == 1 and isinstance(diagnostics[0], CannotOpenFile)
```
madelinezec, line 206: I want to confirm that this should be a CannotOpenFile error, which is what I got when I ran the tests and not a InvalidLiteralInclude like the above asserts?

snooty/types.py
```diff
@@ -466,7 +390,9 @@ def read(
         if match_found:
             for match in match_found:
                 lineno = text.count("\n", 0, match.start())
-                diagnostics.append(Diagnostic.error("git merge conflict found", lineno))
+                diagnostics.append(
+                    GitMergeConflictArtifactFound("git merge conflict found", lineno)
```
i80and, line 394: This is always initialized with the same message, so let's factor this out as well

snooty/parser.py
```diff
@@ -462,7 +456,13 @@ def handle_directive(
                 self.pending.append(PendingFigure(doc, static_asset))
             except OSError as err:
                 msg = f'"{name}" could not open "{argument_text}": {os.strerror(err.errno)}'
-                self.diagnostics.append(CannotOpenFile(msg, util.get_line(node)))
+                # parent_ID = parent_identifier.file
+                # my_File_id = self.asset.fileid
+                self.diagnostics.append(
+                    CannotOpenFile(
+                        argument_text, os.strerror(err.errno), util.get_line(node)
```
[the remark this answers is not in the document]
madelinezec: It is available here. Should I change everywhere I have `os.strerror(err.errno)` to `err.strerror` ?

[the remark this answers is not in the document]
i80and: I would prefer that, yup!

snooty/diagnostics.py
```diff
@@ -0,0 +1,204 @@
...
+
+class UnexpectedIndentation(Diagnostic):
+    severity = Diagnostic.Level.error
+
+
+class InvalidURL(Diagnostic):
+    severity = Diagnostic.Level.error
+
+
+# this could captuer 'literal include expected path arg..as well?
```
sophstad, line 58: [nit] Did you decide whether this will capture the case for literal includes as well? Should probably remove comment either way!

snooty/diagnostics.py
```diff
@@ -0,0 +1,204 @@
...
+    ) -> None:
+        super().__init__(f'"{name}" expected a path argument', start, end)
+        self.name = name
+
+
+class ExpectedImgArg(Diagnostic):
+    severity = Diagnostic.Level.error
+
+
+class ImgExpectedButNotRequired(Diagnostic):
```
sophstad, line 76: [nit] Just because this class name is a bit long, do you think `ImgSuggested` would make sense? I think either way is fine though!

snooty/diagnostics.py
```diff
@@ -0,0 +1,204 @@
...
+
+class InvalidLiteralInclude(Diagnostic):
+    severity = Diagnostic.Level.error
+
+
+class SubstitutionRefError(Diagnostic):
+    severity = Diagnostic.Level.error
+
+
+class VariableNotDeclaredConstant(Diagnostic):
```
sophstad, line 128: [nit] Similarly, we could shorten this class by calling it `ConstantNotDeclared`, but either way 😄

snooty/diagnostics.py
```diff
@@ -0,0 +1,204 @@
...
+
+class AmbiguousTarget(Diagnostic):
+    severity = Diagnostic.Level.error
+
+
+class TodoInfo(Diagnostic):
+    severity = Diagnostic.Level.info
+
+
+class ErrorLoadingFile(Diagnostic):
```
sophstad, line 177: Sorry for not catching this in the spec, but what is the semantic difference between `ErrorLoadingFile` and `CannotOpenFile`? I can see that `ErrorLoadingFile` is specifically thrown when a `LoadError` is caught and `CannotOpenFile` for `OSError` and others, but perhaps there is no meaningful difference to the user and they're worth combining.

snooty/parser.py
```diff
@@ -287,9 +307,7 @@ def dispatch_visit(self, node: docutils.nodes.Node) -> None:
             ), f"Too many ids in this node: {self.docpath} {node}"
             if not node["ids"]:
                 self.diagnostics.append(
-                    Diagnostic.error(
-                        "Links must provide a valid URL", util.get_line(node)
-                    )
+                    InvalidURL("Links must provide a valid URL", util.get_line(node))
```
sophstad, line 310: Any reason why this message is not attached to the class?

snooty/parser.py
```diff
@@ -544,7 +556,7 @@ def handle_directive(
             if not image_argument:
                 # Warn writers that an image is suggested, but do not require
                 self.diagnostics.append(
-                    Diagnostic.warning(
+                    ImgExpectedButNotRequired(
                         f'"{name}" expected an image argument', util.get_line(node)
```
sophstad, line 560: Same here; can we attach the message to the class?

snooty/postprocess.py
```diff
@@ -87,7 +87,7 @@ def __call__(self, filename: FileId, node: n.Node) -> None:
             if not self.pending_program:
                 line = node.start[0]
                 self.diagnostics[filename].append(
-                    Diagnostic.error("'.. option::' must follow '.. program::'", line)
+                    MissingOption("'.. option::' must follow '.. program::'", line)
```
sophstad, line 90: Same

snooty/postprocess.py
```diff
@@ -219,14 +219,14 @@ def handle_refs(self, filename: FileId, node: n.Node) -> None:
         if not target_candidates:
             line = node.span[0]
             self.diagnostics[filename].append(
-                Diagnostic.error(f'Target not found: "{node.name}:{node.target}"', line)
+                TargetNotFound(f'Target not found: "{node.name}:{node.target}"', line)
```
sophstad, line 222: Same

snooty/postprocess.py
```diff
@@ -219,14 +219,14 @@ def handle_refs(self, filename: FileId, node: n.Node) -> None:
...
             )
             return
 
         if len(target_candidates) > 1:
             line = node.span[0]
             self.diagnostics[filename].append(
-                Diagnostic.error(f'Ambiguous target: "{node.name}:{node.target}"', line)
+                AmbiguousTarget(f'Ambiguous target: "{node.name}:{node.target}"', line)
```
sophstad, line 229: Same

snooty/types.py
```diff
@@ -495,7 +419,7 @@ def handle_match(match: Match[str]) -> str:
             except KeyError:
                 lineno = source.count("\n", 0, match.start())
                 diagnostics.append(
-                    Diagnostic.error(
+                    VariableNotDeclaredConstant(
                         f"{variable_name} not defined as a source constant", lineno
```
sophstad, line 423: Same

snooty/diagnostics.py
i80and: Shorter is good! Perhaps `ImageSuggested`?

Sophie, you wrote the logic where this is raised originally, right? If it really is just a suggestion, should it be info rather than warning?

snooty/diagnostics.py
i80and: Ah, actually good call: that's a better name!

snooty/diagnostics.py
madelinezec: I believe it was your call @i80and to have ErrorLoadingFile and then CannotOpenFile and for the to latter capture `CannotOpenFile`,  `FileDoesNotExist`, `CannotOpenLiteralInclude`, `CannotOpenImg`. Do you see any reason to not group `ErrorLoadingFile` into `CannotOpenFile` as well?

snooty/parser.py
madelinezec: I originally only attached message construction logic to the classes where the message was invariable and there was more than one instance of that error (because I was mostly concerned with DRY principle). But after reading your comment and checking in with Andrew, I will now go ahead and attach the message to classes where the message is invariable :)

snooty/diagnostics.py
sophstad: Yup, info sounds good! And if we go with `ImageSuggested`, we should probably change `ExpectedImgArg` to `ExpectedImageArg` for consistency.

madelinezec: Ah good catch

snooty/diagnostics.py
madelinezec: ok @sophstad so we actually changed the name of this error from `ErrorLoadingFile` to `UnmarshallingError` since the issue has to do with the type in the data not matching the python class we are loading it into

sophstad:
Just one note: I merged in some files the other day that include diagnostics. Doesn't look like they are included in this update, so maybe they should be added! See [here](https://github.com/mongodb/snooty-parser/blob/master/snooty/postprocess.py#L207-L228)

PR #134, opened 2020-04-02
DOP-922: Report diagnostics with a separate return code

i80and:
* Return status code 2 if there are build errors
* Status code 1 now indicates a crash
* Fixed up and refactored our diagnostic levels

### DIFF
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -29,15 +29,14 @@ class Diagnostic:
         self.end = (end_line, end_column)
 
     class Level(enum.IntEnum):
-        info = 0
-        error = 1
+        info = 1
         warning = 2
+        error = 3
 
         @classmethod
-        def from_docutils(cls, docutils_level: int) -> "Diagnostic.Level":
-            level = docutils_level - 1
-            level = min(level, cls.warning)
+        def from_docutils(cls, level: int) -> "Diagnostic.Level":
             level = max(level, cls.info)
+            level = min(level, cls.error)
             return cls(level)
 
     @property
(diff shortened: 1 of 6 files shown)


snooty/main.py
```diff
@@ -275,4 +282,4 @@ def main() -> None:
             connection.close()
 
     if args["build"] and backend.total_warnings > 0:
```
skerschb, line 284: Do we want to make this more of a threshold?

snooty/main.py
```diff
@@ -284,5 +288,5 @@ def main() -> None:
             print("Closing connection...")
             connection.close()
 
-    if args["build"] and backend.total_warnings > 0:
-        sys.exit(1)
+    if args["build"] and backend.total_errors > 0:
+        sys.exit(EXIT_STATUS_ERROR_DIAGNOSTICS)
```
madelinezec, line 292: So what is the difference between exit code 1 and 2 here? How does returning exit code 2 make it so the autobuilder can catch unexpected parser terminations?

i80and: Exit code 1 indicates unexpected termination, while 2 indicates that the parse finished, but with errors the writer should definitely look at. If error code 1 is returned, then the autobuilder should not run the frontend at all.

snooty/diagnostics.py
```diff
@@ -29,15 +29,14 @@ def __init__(
         self.end = (end_line, end_column)
 
     class Level(enum.IntEnum):
-        info = 0
-        error = 1
+        info = 1
         warning = 2
+        error = 3
```
mango-db, line 34: Is there a reason why the internal `Level` numbering is the opposite of standard (`error=1`, etc...)?

i80and: This matches the internal doctutils diagnostic code scheme, but that's mostly an arbitrary decision! It is however more common for increasingly severe error conditions to have higher values (see: Python, Java standard libraries). Language Server Protocol is the odd one out, as far as I'm aware.

PR #141, opened 2020-04-08
DOP-908: Parser support for deprecated versions

mango-db:
[DOP 908](https://jira.mongodb.org/browse/DOP-908).


test_data/test_postprocessor/snooty.toml
```diff
@@ -11,3 +11,7 @@ global-write = "Global Writes"
 global-write-clusters = "Global *Clusters*"
 service = "Atlas"
 blank = ""
+
+[deprecated_versions]
+manual = ["v2.2", "v2.4", "v2.6", "v3.0", "v3.2", "v3.4"]
+mms = ["v1.1", "v1.2", "v1.3"]
```
sophstad, line 17: Since you're adding this table to the test project, we should be verifying somewhere that it is actually being ingested and saved properly.

snooty/test_types.py
```diff
@@ -23,6 +23,7 @@ def test_project() -> None:
     project_config, project_diagnostics = ProjectConfig.open(Path(".").resolve())
     assert project_config.name == "unnamed"
     assert project_config.title == "untitled"
+    assert project_config.deprecated_versions == {}
```
sophstad, line 26: Since we have a fairly limited use case for `deprecated_versions` data (only in docs-landing), I think I would prefer to not write that field to the metadata collection at all if nothing is specified. Do you think that makes sense?

snooty/postprocess.py
```diff
@@ -170,6 +170,7 @@ def run(
     def generate_metadata(self) -> n.SerializedNode:
         document: Dict[str, SerializableType] = {}
         document["title"] = self.project_config.title
+        document["deprecated_versions"] = self.project_config.deprecated_versions
```
sophstad, line 173: When I parsed a repo that doesn't have a `deprecated_versions` table in its snooty.toml, the `metadata` document contained the line `deprecated_versions: null`. We need to wrap this in a conditional so that it's only included if that field is defined:
```python
if self.project_config.deprecated_versions:
    document["deprecated_versions"] = self.project_config.deprecated_versions
```

PR #142, opened 2020-04-08
DOP-954: Attach :doc: titles in postprocessor

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOP-954)]
- Convert `:doc:` to `ref_role` in `JSONVisitor` class
  - RefRoles with `name=doc` do not contain targets. We use targets on the front-end to navigate to a particular id (e.g. `/fileid#target`), so omitting this navigates to the root doc, as is expected behavior with `:doc:` roles. Let me know if this doesn't make sense!
  - Doc refroles contain `fileid` properties to navigate to the correct page

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -2,6 +2,7 @@ import os.path
 import logging
 from collections import defaultdict
 from copy import deepcopy
+from pathlib import PurePath
 from typing import (
     Any,
     Callable,
@@ -23,7 +24,7 @@ from .types import (
     SerializableType,
     TargetDatabase,
 )
-from . import n
+from . import n, util
 from .util import SOURCE_FILE_EXTENSIONS
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


snooty/postprocess.py
```diff
@@ -204,6 +205,40 @@ def run_event_parser(
...
+        title = self.slug_title_mapping.get(slug)
+
+        if not title:
+            line = node.span[0]
+            self.diagnostics[filename].append(
+                Diagnostic.error(f"Page title not found: {target_path}", line)
+            )
+            return
+
+        page_title_nodes: MutableSequence[n.InlineNode] = list(
```
i80and, line 235: You can just do `= [deepcopy(node) for node in title]`

snooty/postprocess.py
```diff
@@ -204,6 +205,40 @@ def run_event_parser(
...
+    def _attach_doc_title(self, filename: FileId, node: n.RefRole) -> None:
+        if not isinstance(node.fileid, str):
+            line = node.span[0]
+            self.diagnostics[filename].append(
+                Diagnostic.error(f'File not specified: "{node.name}"', line)
+            )
+            return
+
+        # Add .txt to end of doc role target path
+        target_path = PurePath(node.fileid)
```
i80and, line 217: Hm. A couple questions:

* Why are we instantiating a PurePath rather than keeping it as a FileId?
* Do we need to append the `.txt` suffix, since that's not actually used in the lookup afaics?

sophstad: Good point! I did this because I wasn't sure if `reroot_path()` would work properly without the extensions, but... it does! Thanks 🎉

snooty/postprocess.py
```diff
@@ -213,31 +212,20 @@ def _attach_doc_title(self, filename: FileId, node: n.RefRole) -> None:
...
 
-        # Add .txt to end of doc role target path
-        target_path = PurePath(node.fileid)
-        # Adding the current suffix first takes into account dotted targets
-        new_suffix = target_path.suffix + ".txt"
-        target_path = target_path.with_suffix(new_suffix)
-
         relative, _ = util.reroot_path(
-            target_path, filename, self.project_config.source_path
+            FileId(node.fileid), filename, self.project_config.source_path
```
i80and, line 216: fileid is already a FileId, right? Nice thing about FileId is that it's immutable, so you never need to copy it for any purpose!

PR #147, opened 2020-04-13
Tweak :doc: role

i80and:
* Place `doc` in the `std` domain
* For each page top-level heading, create a target


i80and:
> lgtm! Just one question: do you envision `:doc:` roles linking directly to the heading (i.e. to `/page#heading-id`)? Currently they link directly to the page, no ID included.

I actually would prefer the role to link to the page, rather than the first heading, personally, breaking with Sphinx! Do you have a preference?

PR #152, opened 2020-04-22
DOP 888: Add glossary support

madelinezec:


snooty/test_parser.py
```diff
@@ -612,6 +612,92 @@ def test_accidental_indentation() -> None:
...
+        page.ast,
+        """
+<root> 
+  <directive name="glossary" sorted="True">
+    <definitionList>       
+      <definitionListItem>         
+        <term>           
+          <text>$cmd</text>
+          <inline_target domain="std" name="term">            
+            <target_identifier ids="['cmd']">
```
madelinezec, line 654: I know in the jira ticket you had the id as `$cmd` but I looked into docutils make_id and the docs say, "ID and NAME tokens must begin with a letter ([A-Za-z]) and may be followed by any number of letters, digits ([0-9]), hyphens (“-”), underscores (“_”), colons (”:”), and periods (”.”)." ([source](ID and NAME tokens must begin with a letter ([A-Za-z]) and may be followed by any number of letters, digits ([0-9]), hyphens (“-”), underscores (“_”), colons (”:”), and periods (”.”).)

snooty/test_parser.py
```diff
@@ -612,6 +612,92 @@ def test_accidental_indentation() -> None:
...
+          </inline_target>         
+        </term>
+        <paragraph><text>foobar</text></paragraph>       
+      </definitionListItem>        
+
+      <definitionListItem>
+        <term>           
+          <text>_id</text>           
+          <inline_target domain="std" name="term">
+            <target_identifier ids="['id']">
```
madelinezec, line 665: why the target id is not _id: "For CSS compatibility, identifiers (the “class” and “id” attributes) should have no underscores, colons, or periods. Hyphens may be used." ([docutils docs])(http://code.nabla.net/doc/docutils/api/docutils/nodes/docutils.nodes.make_id.html)

i80and: :+1:
(Note that while we currently use `make_id()`, I don't consider that a long-term solution, and I'll re-evaluate in DOP-998)

snooty/parser.py
```diff
@@ -433,6 +435,34 @@ def dispatch_departure(self, node: docutils.nodes.Node) -> None:
...
                 )
 
+        if (
+            isinstance(popped, n.Directive)
+            and f"{popped.domain}:{popped.name}" == ":glossary"
+        ):
+
+            definition_list = next(popped.get_child_of_type(n.DefinitionList), None)
+
+            if definition_list is None:
```
i80and, line 445: Our tests should take this branch.

It might also be worth handling the case where there's more than one child node? if len(popped.children) != 1

madelinezec: if a glossary has more than one child node - does that mean it has two separate definition lists children? or it has two or more `definition_list_item` children without a `definition_list parent container`?

snooty/diagnostics.py
```diff
@@ -194,6 +194,21 @@ def __init__(
...
+class MalformedGlossary(Diagnostic):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(
+            f"Malformed glossary: glossary is either missing a definition list or has more than one child node",
```
i80and, line 206: `child node` is too jargony: writers don't know or care about what a node is

PR #157, opened 2020-05-11
DOP-876: populate literal include nodes

mango-db:
## Staging
- [Docs-node](https://docs-mongodbcom-staging.corp.mongodb.com/master/node/claire/master/)

## Notes
- All handing of code-blocks was moved from `parser.py` (`PendingLiteralInclude` class, `PendingFigure.handle_directive()`) to `postprocess.py`
- Due to the logic moving, some moderate reductions were made to the tests available in  `test_parser.py` and `test_project.py` 
- All literalinclude tests were moved to `test_postprocess.py`, accompanied by new files` literal-include.txt`, `sample_code.js` and `sample_code.py`
- Semi-related `test_expand_includes` was modified to conform to the new `check_ast_testing_string` norm
- We no longer confirm that modifying a literally-included asset reparses the dependent files


snooty/postprocess.py
```diff
@@ -413,6 +416,94 @@ def get_include_argument(node: n.Directive) -> str:
...
+            # Attempt to read the literally included file
+            try:
+                with open(include_filepath) as file:
+                    text = file.read()
+
+            except OSError as err:
+                self.diagnostics[filename].append(
+                    CannotOpenFile(include_filepath, err.strerror, node.span + (-1,))
+                )
+                return
```
sophstad, line 435: The old `PendingLiteralInclude` implementation uses this [`add_static_asset()`](https://github.com/mongodb/snooty-parser/blob/master/snooty/parser.py#L665) function, which takes care of generating the correct path and loading the file in a safe manner. Perhaps you can try using `reroot_path()` and `StaticAsset.load()` here to leverage some of our existing code!

snooty/postprocess.py
```diff
@@ -413,6 +416,94 @@ def get_include_argument(node: n.Directive) -> str:
...
+                        f'"{end_before_text}" precedes {start_after_text} in {include_filepath}',
+                        node.span + (-1,),
+                    )
+                )
+                return
+
+            lines = lines[start_after:end_before]
+
+            if "dedent" in node.options:
+                assert isinstance(node.options["dedent"], int)
```
sophstad, line 496: `dedent` can be an integer _or_ a flag according to the [rstspec.toml](https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml#L118), so this assertion is invalid. We should also add some handling for the flag variant, in which the dedent amount is determined automatically (as was done in `PendingLiteralInclude`).

snooty/test_postprocess.py
```diff
@@ -47,76 +52,87 @@ def test_deprecated_versions(backend: Backend) -> None:
...
+    # Test failure to locate end-before text
+    literal_include_node = ast.children[4]
+    assert isinstance(diagnostics[2], InvalidLiteralInclude)
+
+    # Test start-after text is below end-before text
+    literal_include_node = ast.children[5]
+    assert isinstance(diagnostics[3], InvalidLiteralInclude)
+
+
 def test_expand_includes(backend: Backend) -> None:
```
sophstad, line 107: Again, thanks for fixing this test 💯

snooty/test_postprocess.py
```diff
@@ -47,76 +52,87 @@ def test_deprecated_versions(backend: Backend) -> None:
...
+</directive>""",
+    )
+
+    # Test a literally-included code block with fully specified options
+    literal_include_node = ast.children[1]
+    check_ast_testing_string(
+        literal_include_node,
+        """<directive name="literalinclude" dedent="4" end-before="end example 1" language="python" start-after="start example 1">
+        <text>includes/sample_code.py</text>
+        <code copyable="True" lang="python" linenos="True">print("test dedent")</code>
```
sophstad, line 82: `copyable` and `linenos` were not specified in the `literalinclude` usage, so they actually shouldn't be appearing here! This is related to my comment above about the instantiation of `Code`.

mango-db: Since `copyable` [defaults to True](https://github.com/mongodb/snooty-parser/blob/master/snooty/parser.py#L546), I think it should show up in the default example, no? But `linenos`, I agree

snooty/postprocess.py
```diff
@@ -413,6 +416,94 @@ def get_include_argument(node: n.Directive) -> str:
...
+                lines = [line[node.options["dedent"] :] for line in lines]
+
+            selected_content = "\n".join(lines)
+
+            language = node.options["language"] if "language" in node.options else ""
+
+            code = n.Code((1,), language, True, [], selected_content, True)
```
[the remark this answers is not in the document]
mango-db: So, I'm not seeing a `copyable` as a directive option -- [only as a code-block option](https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml#L262). Should I add it as a directive option?

mango-db:
Example of docs-node staging with literally-included code: [crud/read-operations/sort](https://docs-mongodbcom-staging.corp.mongodb.com/master/node/claire/master/fundamentals/crud/read-operations/sort)

snooty/test_parser.py
```diff
@@ -140,73 +141,165 @@ def test_literalinclude() -> None:
...
-.. literalinclude:: /driver-examples/garbagnrekvjisd.py
+.. literalinclude:: /test_parser/includes/sample_code.py
+   :start-after: does not exist
+   :end-before: end example 1
 """,
     )
     page.finish(diagnostics)
-    assert len(diagnostics) == 1 and isinstance(diagnostics[0], CannotOpenFile)
+    assert len(diagnostics) == 1
+    assert isinstance(diagnostics[0], InvalidLiteralInclude)
```
sophstad, line 243: I like the handling of this error a lot—acting as if there is no `start-after` specified and still applying the rest of the options 👍 Perfect!

snooty/test_parser.py
```diff
@@ -140,73 +141,165 @@ def test_literalinclude() -> None:
...
+    check_ast_testing_string(
+        page.ast,
+        """<root>
+        <directive name="literalinclude" start-after="start example 1" end-before="does not exist">
+        <text>/test_parser/includes/sample_code.py</text>
+        <code copyable="True">    print("test dedent")
+    # end example 1
+
+    # start example 2
+    print("hello world")</code></directive>
```
sophstad, line 277: Hmm why isn't the line `# end example 2` included here? Is it getting cut off for some reason?

snooty/test_postprocess.py
```diff
@@ -47,76 +53,91 @@ def test_deprecated_versions(backend: Backend) -> None:
     }
 
 
+def test_literal_includes(backend: Backend) -> None:
```
sophstad, line 56: Any reason this is still included here? Feel like we can handle all literal include testing in `test_parser.py` since the postprocessor isn't actually doing anything.

snooty/test_project.py
```diff
@@ -63,17 +63,19 @@ def test() -> None:
...
-        assert len(index.static_assets) == 2
+        assert len(index.static_assets) == 1
         assert not index.pending_tasks
         for node in ast_dive(index.ast):
             if isinstance(node, n.Code):
                 code_length += len(node.value)
             elif isinstance(node, n.Directive) and node.name == "figure":
                 checksums.append(node.options["checksum"])
-        assert code_length == 345
+        assert code_length == 400
```
sophstad, line 78: What's the reason for this change? Is it because of the accidental removal of line 0 that you fixed?

snooty/util.py
```diff
@@ -192,13 +192,21 @@ def __len__(self) -> int:
         return sum(len(w) for w in self.directories.values())
 
 
+# Should this really default to flag-like behavior? Seems iffy at best
```
sophstad, line 195: I think changing this behavior could have pretty wide-ranging implications across repositories! There are lots of different boolean options and we want to preserve the usage expected by docutils, and yield the same output. Did you change this because you were running in to problems?

snooty/parser.py
```diff
@@ -531,35 +433,115 @@ def handle_directive(
...
+            copyable = "copyable" not in options or options["copyable"] == "True"
+            emphasize_lines = (
+                rstparser.parse_linenos(options["emphasize-lines"], len(lines))
+                if "emphasize-lines" in options
+                else None
+            )
+            selected_content = "\n".join(lines)
+            linenos = "linenos" in options
+
+            # Code(span, language, copyable, emphasize_lines, value, linenos)
```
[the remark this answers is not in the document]
mango-db: I assume this is referencing the comment `# Code(span, language, copyable, emphasize_lines, value, linenos)`, and am resolving the conversation.

snooty/test_parser.py
mango-db: I've updated `end_before` functionality to "match". If no `end_before_text` is found, it defaults to the EOF.

snooty/test_parser.py
mango-db: The original functionality would have cut off the last line (due to setting `end_before` to `-1`, then splicing `lines`). I've updated it such that if no `end_before` is found, it defaults to EOF (as if no `end_before` was specified).

snooty/test_postprocess.py
mango-db: Yeah, I've also considered deleting it. I've left it in since it's pretty concise, but it's up to you.

snooty/test_project.py
mango-db: The length of `index.static_assets` drops by one, because our literally included file (`test_project/source/driver-examples/DocumentationExamples.cs`) is [no longer a static_asset](https://github.com/mongodb/snooty-parser/blob/master/snooty/parser.py#L553). I don't believe that `code_length` increases because of that `start_after` OBOE, since the `index.txt` specifies the `start_after` text. I'll look into this.

snooty/util.py
mango-db: With regards to the function, I am indeed running into issues where `:copyable: false` and `:copyable: False` both output `True`. I've temporarily edited this function for the commit -- we should definitely chat about the line `docutils.parsers.rst.directives.choice(argument, ("true", "false"))`.

snooty/test_postprocess.py
sophstad: I don't think it should be included since we don't touch literal includes in the postprocess layer!

snooty/test_project.py
mango-db: It was due to a misconfigured `dedent` conditional, and has been fixed!

mango-db:
Update staging:
- [Docs-node](https://docs-mongodborg-staging.corp.mongodb.com/master/node/claire/master/
)
- [Docs-ecosystem](https://docs-mongodborg-staging.corp.mongodb.com/master/ecosystem/claire/master/)
- [Docs-realm](https://docs-mongodborg-staging.corp.mongodb.com/master/realm/claire/master/)
- [Docs-tutorial](https://docs-mongodborg-staging.corp.mongodb.com/master/guides/claire/master/
)

snooty/parser.py
```diff
@@ -534,32 +436,120 @@ def handle_directive(
...
+                end_before = len(lines)
+
+            # Capture the original file-length before splicing it
+            len_file = len(lines)
+            lines = lines[start_after:end_before]
+
+            dedent = 0
+            if "dedent" in options:
+                # Dedent is specified as a nonnegative integer (number of characters)
+                if type(options["dedent"]) is int:
```
sophstad, line 503: [nit] What about using `isinstance`?
```suggestion
                if isinstance(options["dedent"], int):
```

mango-db: This is intentional! Since `bool` is a subtype of `int`, and since `options["dedent"]` can take either type, there was an error in which this line was catching both types. I'll just swap the order instead, in order to preserve the `isinstance` call.

PR #163, opened 2020-05-27
DOP-1128: adds path to rfc role to facilitate use

schmalliso:


snooty/rstspec.toml
```diff
@@ -711,7 +711,7 @@ type = "text"
 
 [role.rfc]
 help = """Reference an IETF RFC."""
-type = "explicit_title"
+type = {link = "https://https://tools.ietf.org/html/%s"}
```
i80and, line 714: `https://https://`? :D

schmalliso: oh ffs.

schmalliso: Updated.

snooty/rstspec.toml
```diff
@@ -711,7 +711,7 @@ type = "text"
 
 [role.rfc]
 help = """Reference an IETF RFC."""
-type = "explicit_title"
+type = {link = "https://tools.ietf.org/html/%s"}
```
i80and, line 714: Apparently IETF does have a redirect set up, but technically there should be an `rfc` prefix; i.e.
`/html/rfc%s` e.g. https://tools.ietf.org/html/rfc7919#appendix-A.2

PR #168, opened 2020-06-15
DOP-1073: no longer render RefRole prefix if no title is found

i80and:
Much refactoring was required to avoid circular import errors. This improves the overall code modularity, but makes this a bit bloated. Sorry!

Additionally, the spec is now a global, and is parsed upon specparser module import.

This is the resulting change in the output AST: https://github.com/mongodb/snooty-parser/pull/168/files#diff-915b6e0f0f3fa5b6662ec3cdccbee909R143

The specific example Jeff encountered is now output as:
```
{
    "type": "ref_role",
    "position": {
        "start": {
            "line": 107
        }
    },
    "children": [
        {
            "type": "literal",
            "position": {
                "start": {
                    "line": 107
                }
            },
            "children": [
                {
                    "type": "text",
                    "position": {
                        "start": {
                            "line": 107
                        }
                    },
                    "value": "$match"
                }
            ]
        }
    ],
    "domain": "mongodb",
    "name": "pipeline",
    "target": "pipe.$match",
    "flag": "",
    "url": "https://docs.mongodb.com/manual/reference/operator/aggregation/match/#pipe._S_match"
}
```

### DIFF
--- a/snooty/intersphinx.py
+++ b/snooty/intersphinx.py
@@ -29,7 +29,11 @@ class TargetDefinition(NamedTuple):
     priority: int
     uri_base: str
     uri: str
-    display_name: str
+    display_name: Optional[str]
+
+    @property
+    def domain_and_role(self) -> str:
+        return f"{self.role[0]}:{self.role[1]}"
 
 
 @dataclass
@@ -73,9 +77,7 @@ class Inventory:
                         ":".join(target.role),
                         str(target.priority),
                         target.uri_base,
-                        "-"
-                        if target.name == target.display_name
-                        else target.display_name,
+                        "-" if target.display_name is None else target.display_name,
                     )
                 )
             )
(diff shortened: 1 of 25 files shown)


snooty/intersphinx.py
```diff
@@ -127,8 +129,7 @@ def parse(cls, base_url: str, text: bytes) -> "Inventory":
             domain, role = domain_and_role.split(":", 1)
 
             # "If {dispname} is identical to {name}, it is stored as -"
-            if dispname == "-":
-                dispname = name
+            dispname = None if raw_dispname == "-" else raw_dispname
```
sophstad, line 132: Just wondering, where does the hyphen come from?

i80and: I dug in quickly to ensure that I had this right, but sphinx uses a hyphen indicates that the label should be the same as the target name. For... some reason.

PR #169, opened 2020-06-16
Dop 1132

madelinezec:
This PR contains the changes necessary for implementing the fail_on_diagnostics flag. You can check out the ticket here: https://jira.mongodb.org/browse/DOP-1132

This works by adding the `fail_on_diagnostics` flag to a docs repo `snooty.toml` and setting it to a boolean. If true, the parser will throw a error with exit code 1. The autobuilder maintained makefiles will catch this failure and will report it in the logs of the build. 

This PR also contains a test in test_main.py for testing this parser failure.


mango-db:
Small q, when I run `make lint`, I'm getting a number of `snooty/test_parser.py:168: error: Missing positional argument "fail_on_diagnostics" in call to "ProjectConfig"` and `snooty/test_parser.py:62: error: Too few arguments for "ProjectConfig"` errors. Is that intentional?

Also, be sure to add your change to `CHANGELOG.md`!

madelinezec:
> Small q, when I run `make lint`, I'm getting a number of `snooty/test_parser.py:168: error: Missing positional argument "fail_on_diagnostics" in call to "ProjectConfig"` and `snooty/test_parser.py:62: error: Too few arguments for "ProjectConfig"` errors. Is that intentional?
> 
> Also, be sure to add your change to `CHANGELOG.md`!

Ah great catch!! That is not intentional - I had to add something to the ProjectConfig object so now all the tests that rely on that class are going to fail. I'll fix that - thanks for catching it :)

snooty/main.py
```diff
@@ -249,9 +249,9 @@ def _generate_build_identifiers(args: Dict[str, Optional[str]]) -> BuildIdentifi
     return identifiers
 
 
-def main() -> None:
+def main(argv: Optional[Sequence[str]] = None) -> None:
```
sophstad, line 252: Sorry, was snooping and just wondering: what's the point of adding this `argv` variable? Is there a way to set `fail_on_diagnostics` from the command line? If not, it seems like `argv` isn't actually used anywhere and this work is mostly a `ProjectConfig` change.

i80and: oh, interesting. This was used when main was going to be invoked internally by the test, but it looks like it's now called via subprocess.

snooty/util.py
```diff
@@ -50,7 +50,7 @@ def get_files(root: PurePath, extensions: Container[str]) -> Iterator[Path]:
     for base, dirs, files in os.walk(root):
         for name in files:
             ext = os.path.splitext(name)[1]
-
+            print(base, " ", ext)
```
i80and, line 53: We definitely don't want to be printing in a utility function like this

snooty/types.py
```diff
@@ -413,7 +414,7 @@ def open(cls, root: Path) -> Tuple["ProjectConfig", List[Diagnostic]]:
 
             path = path.parent
 
-        return cls(root, name="unnamed"), diagnostics
+        return cls(root, name="unnamed", fail_on_diagnostics=False), diagnostics
```
i80and, line 417: You shouldn't need to manually specify this, because it's a default field

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- fail_on_diagnostics toml flag
```
i80and, line 12: Best to include the jira ticket, as seen in the other lines.

Also you can monospace `fail_on_diagnostics` since this is markdown, and specify `snooty.toml`, since there's a number of toml files that snooty ingests.

snooty/util.py
madelinezec: ack! yes

PR #172, opened 2020-06-22
DOP-1148: Improve string-handling

mango-db:
#### `rstspec.toml`
- Change `options.start-after` and `options.end-before` to `string`s to accurately reflect their contents (text, not line numbers). (Original type `linenos` were converting to strings anyways in `specparser.py` `PrimitiveType Validators`, so this is functionally the same.)
#### `specparser.py`
- `specparser.py`'s `path`, `string`, and `linenos` `PrimitiveType`s are passed to `util.option_string` instead of `str` function. 
#### `util.py`
- Add a new string-handling function `util.option_string`. This precludes the conversion of `NoneType None` to a `StringType "None"`, which incorrectly caused the parser to search documents for `"None"` `start_after_text` and `end_before_text` when writers fail to specify options


snooty/util.py
```diff
@@ -192,6 +192,16 @@ def __len__(self) -> int:
...
 
 
+def option_string(argument: Optional[str]) -> Optional[str]:
+    """
+    Check for a valid string option and return it. If no argument is given,
+    return None.
+    """
+    if argument and argument.strip():
+        return argument
+    return None
```
sophstad, line 202: Let's raise a diagnostic before returning `None` to warn writers that they're using the option incorrectly!

i80and: Specifically, I think your only option is to raise ValueError

mango-db: I've added a small check to parser.py instead, so that we don't have to raise a ValueError (which might break current builds). Instead, I added a diagnostic for start-after and end-before.

mango-db:
Updated staging:

[Docs-node](https://docs-mongodborg-staging.corp.mongodb.com/master/node/claire/master/
)

snooty/util.py
mango-db: OK, so I did end up adding the `raise ValueError`, per Andrew's suggestion. Corresponding tests are in `test_util.py`.

PR #177, opened 2020-06-29
Dop 1171

madelinezec:
https://jira.mongodb.org/browse/DOP-1171 

prepend `term` to the term's name in the glossary node


snooty/parser.py
```diff
@@ -363,7 +363,7 @@ def dispatch_departure(self, node: docutils.nodes.Node) -> None:
 
             for item in definition_list.get_child_of_type(n.DefinitionListItem):
                 term_text = "".join(term.get_text() for term in item.term)
-                term_identifier = make_id(term_text)
+                term_identifier = make_id("term-" + term_text)
```
sophstad, line 366: [nit] This would be a good place to use an [f-string](https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/)!
```suggestion
                term_identifier = make_id(f"term-{term_text}")
```

madelinezec:
> Thanks for this fix! Just a tiny suggestion 😄

nice! committed

PR #179, opened 2020-07-09
Dop 1188: fix for labels starting with number

madelinezec:
We discovered an important docutils distinction: 

> Names are case- and whitespace-normalized (see the
> fully_normalize_name() function), and IDs conform to the regular
> expression ``[a-z](-?[a-z0-9]+)*`` (see the make_id() function).

By changing the snooty parser ref role node's id to equal the docutils node's name (instead of `id`) in `snooty/parser.py`, refroles correctly have fileids! 

see this ticket https://jira.mongodb.org/browse/DOP-1188 and allison's [staging link](https://docs-mongodbcom-staging.corp.mongodb.com/charts/docsworker-xlarge/digit-ref-example/chart-type-reference/column-bar-chart) that demonstrates the issue


madelinezec:
> A staging link demonstrating functionality would be helpful! Here's mine: [Docs-charts](https://docs-mongodbcom-staging.corp.mongodb.com/master/charts/claire/DOP-1130/chart-type-reference/column-bar-chart#100-stacked-example). It seems like this is in-fact [correctly using a label](https://github.com/10gen/docs-charts/blob/master/source/chart-type-reference/column-bar-chart.txt#L113) `<100-stacked-example>` starting with a number in the docs, if I understand labels correctly. Can you confirm that to me?

I'm not sure we do staging links for parser PRs ? At least I've never been by Andrew for that before, but it is an indea. But the ticket linked at the top of this PR that Allison created does have the necessary info because it was a bug she found :)

snooty/test_parser.py
```diff
@@ -534,6 +534,36 @@ def test_rst_replacement() -> None:
...
+    parser = rstparser.Parser(project_config, JSONVisitor)
+    
+    # test label starting with a number
+    page, diagnostics = parse_rst(
+        parser,
+        path,
+        """
+:ref:`100_stacked_example`
+
+.. _100_stacked_example:
```
i80and, line 549: I forget -- is it possible to test the TargetIdentifier in this node in the parser tests?

snooty/test_postprocess.py
```diff
@@ -143,10 +143,37 @@ def test_validate_ref_targets(backend: Backend) -> None:
         <literal><text>mongod</text></literal>
         </ref_role>""",
     )
+    # Assert that refs beginning with number work
```
i80and, line 146: More broadly, that also contain underscores

snooty/test_postprocess.py
```diff
@@ -143,10 +143,37 @@ def test_validate_ref_targets(backend: Backend) -> None:
...
+        name="label"
+        target="z100_stacked_example"
+        fileid="index">
+        <text>Z100 Stacked Examples</text>
+        </ref_role>""",
+    )
     # Check that undeclared targets raise an error
     diagnostics = backend.diagnostics[page_id]
-    assert len(diagnostics) == 1
+    assert len(diagnostics) == 2
```
i80and, line 176: This test might be fragile: it depends on the order of the diagnostics staying the same, which might not be a guarantee we want to make.

better to test that diagnostics *contains* the right diagnostic types

i80and:
Merged in 20448b40f026c43e6cc26f37797af8778aeb3b1d

PR #181, opened 2020-07-14
DOP-1243: Modify admonition configuration

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOP-1243)]
- Deprecate `admonition`, `caution`, and `topic` admonition directives
- `see`, `seealso`, and `example` inherit the `admonition` directive config
- All admonitions (`note`, `important`, `warning`, `tip`, `see`, `seealso`, `example`) now accept titles because the [LeafyGreen component](https://mongodb.github.io/leafygreen-ui/?path=/story/callout--default) distinguishes between the title and the admonition type. Lemme know if I messed up the snippets!

**Note:** `danger` doesn't appear anywhere in docs content afaict, but it was included in rstspec.toml so I marked it as deprecated. We could delete it instead, though!

### DIFF
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -59,6 +59,7 @@ ${4:   :align:
 content_type = "block"
 argument_type = "string"
 options.class = "string"
+deprecated = true
 example = """.. admonition:: ${1:Title}
    
    ${2:Admonition content}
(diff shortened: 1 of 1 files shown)


snooty/rstspec.toml
```diff
@@ -590,8 +592,8 @@ inherit = "_guides-base"
 
 [directive.seealso]
 help = """Related pages. Typically a list."""
-inherit = "_guides-base"
-# example = """.. seealso::
+inherit = "admonition"
```
schmalliso, line 595: I think this seealso is guides-specific, rather than the sphinx-y `seealso` directive. This one should maybe be scoped to guides only and then have a separate `seealso` that's like the `see`?

sophstad: Good point, there is a `seealso` that's used everywhere so I assumed they were the same. I will add one with the rest of the admonitions and revert this to its previous config.

PR #187, opened 2020-07-29
DOP-1295: Support directive fields

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOP-1295)] Things got a bit messy with the `try`/`except` statements, so would appreciate any ideas you have to clean that up!

Also if you know of any other directive fields to add to the rstspec.toml, let me know!


snooty/parser.py
```diff
@@ -136,24 +136,65 @@ def __init__(
...
+                        None if isinstance(f, str) else f[1]
+                    )
+                    for f in rstobject["fields"]
+                }
+
+                self.state.append(
+                    n.Field((line,), [], field_name, supported_fields[field_name])
+                )
+                return
+            except AssertionError:
```
i80and, line 175: Assertions shouldn't really be used this way: assert statements MAY be turned off at runtime (see: https://docs.python.org/3/using/cmdline.html#cmdoption-o), so they must only be used for truly invalid states where the only safe thing to do is abort.

Can you just use an if statement for this?

PR #189, opened 2020-08-04
DOP-1269: Allow for additional list-table width delimiters

mango-db:
[JIRA.](https://jira.mongodb.org/browse/DOP-1269)

- Allow for additional delimiters for width options per the [rst list-table directive](https://docutils.sourceforge.io/docs/ref/rst/directives.html#list-table).
  - ` `
  - `, `
  -  `  `
- Test for these new options

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- The :list-table: directive no longer generates incorrect warnings
+
 ## [v0.5.3] - 2020-07-29
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- The :list-table: directive no longer generates incorrect warnings
+- The list-table directive no longer generates incorrect warnings
 
 ## [v0.5.3] - 2020-07-29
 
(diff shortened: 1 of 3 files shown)


snooty/parser.py
```diff
@@ -446,7 +447,7 @@ def handle_directive(
             # Calculate the expected number of columns for this list-table structure.
             expected_num_columns = 0
             if "widths" in options:
-                expected_num_columns = len(options["widths"].split(" "))
+                expected_num_columns = len(re.split("[,* *]+", options["widths"]))
```
sophstad, line 450: I believe that this regex would permit multiple commas in a row, which we may want to avoid? In my opinion `:widths: 38,,72` should not be treated the same as `:widths: 38,72`. What do you think?

Additionally [`\s`](https://docs.python.org/3/library/re.html#index-30) may better represent whitespace, as it is a little more forgiving than `' '`

snooty/test_parser.py
```diff
@@ -1206,6 +1209,67 @@ def test_list_table() -> None:
     page.finish(diagnostics)
     assert len(diagnostics) == 0
 
+    # Comma-separated width option should not fail
```
sophstad, line 1212: These new tests look great!

snooty/parser.py
mango-db: Agreed! Updating

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- The :list-table: directive no longer generates incorrect warnings
```
sophstad, line 11: [nit] The surrounding colons indicate that `list-table` is a role, which it isn't. Directive usage would look like this:
```
.. list-table::
```

But it's probably reasonable to omit these decorators altogether in the changelpg!

mango-db: Thanks for catching that -- Done!

PR #190, opened 2020-08-04
DOCSP-11615: update urls for newest versions of Java driver

ccho-mongodb:
JIRA:
https://jira.mongodb.com/browse/DOCSP-11615


ccho-mongodb:
> Could be worth updating the rest of these links to `https` while you're here, but LGTM!

Thanks! Unfortunately, they're hosted on an insecure site. Will bring this up to Java Drivers Team see if they can get a certificate.

PR #192, opened 2020-08-04
DOP-1306: Support required directive options

i80and:


snooty/test_parser.py
```diff
@@ -1634,3 +1634,26 @@ def test_definition_list() -> None:
...
+    project_config = ProjectConfig(ROOT_PATH, "", source="./")
+    parser = rstparser.Parser(project_config, JSONVisitor)
+
+    page, diagnostics = parse_rst(
+        parser,
+        path,
+        """
+.. figure:: compass-create-database.png""",
+    )
+    assert [type(d) for d in diagnostics] == [DocUtilsParseError]
```
sophstad, line 1650: For my own edification: I don't see `DocUtilsParseError` used elsewhere in this PR. How exactly is it thrown?

i80and: It gets bubbled up by
https://github.com/mongodb/snooty-parser/pull/192/files#diff-74eab25ae48233a33ada7e8ec4f72d21R418-R421

I would actually prefer to have a more specific diagnostic type, but for now this is the best we can do without additional infrastructure.

PR #199, opened 2020-08-12
DOP-1354: Improve unknown node type message & fix <classifier> crash

i80and:
This will make figuring out future issues like this easier.

Right now, if docutils does something unexpected, we provide the *name* of the node, but not the context where it was created, which is less than useful.


snooty/test_parser.py
```diff
@@ -1722,6 +1722,32 @@ def test_definition_list() -> None:
 """,
     )
 
+    # Test a bizarre case
+    page, diagnostics = parse_rst(
+        parser,
+        path,
+        """
```
schmalliso, line 1729: I don't actually know how these tests work, so this is for my own edification. Is this in theory demonstrating that the `collection.createIndex` bit was indented to be definition-list-esque?

i80and: Yup! This inputs the specific case you ran into, and tests that:
* It doesn't crash
* It doesn't try to create a weird node

I'd like to be a lot stricter about how indentation works and provide a diagnostic, but I don't know what that would look like 😅

PR #204, opened 2020-08-23
DOP-1279: Provide explicit HTML IDs in AST

i80and:
* Add an `html_id` attribute to the `Target` node.
* Modify the `fileid` attribute of the `RefRole` node to contain a (filename, elementID) tuple.

**Because we generate element IDs with a DIFFERENT algorithm than Sphinx, rebuilding a property to include this change means we MUST rebuild all dependent properties.**

For example, if we rebuild the manual with snooty, we must now rebuild all projects both old-gen and next-gen that depend on the manual's intersphinx inventory.


snooty/test_parser.py
```diff
@@ -876,7 +876,7 @@ def test_glossary_node() -> None:
         <term>
           <text>$cmd</text>
           <inline_target domain="std" name="term">
-            <target_identifier ids="['term-cmd']">
+            <target_identifier ids="['$cmd']">
```
sophstad, line 879: I think it would be problematic for Snooty to generate different IDs than Sphinx tooling for the glossary (Sphinx IDs being `term-{term}`). This is especially relevant given that `mms-docs` may not be migrated. They have 1000+ uses of `:term:` on the master branch alone, which would mean a lot of broken links.

snooty/n.py
```diff
@@ -338,7 +339,7 @@ class Role(InlineParent):
 class RefRole(Role):
     __slots__ = ("fileid", "url")
     type = "ref_role"
-    fileid: Optional[str]
+    fileid: Optional[Tuple[str, Optional[str]]]
```
sophstad, line 342: Just curious about why it's preferable to use a tuple here rather than add an `html_id` field!

snooty/test_parser.py
i80and: I think it should be fine! Targets are exported and imported between properties through intersphinx, which encodes the actual URL. It is probably worth my testing this quickly just to make sure though

i80and: Ah, good point!

i80and: I think this is actually fine!

The actual target ID in Sphinx is `$cmd`, NOT `term-cmd`: that's a bug that currently keeps terms from working.

The element ID is fine to change because it's only exposed through the intersphinx inventory, which contains the full url.

Let me know if I'm missing something!

snooty/n.py
i80and: Yes! So let me know if you think this is silly and unnecessary, but this lets us encode in the type system that the `html_id` can be set if and only if the `fileid` is set! Except, this type signature can actually be changed to `Optional[Tuple[str, str]]`, which would say "fileid and html_id must both be set, or neither are set".

If they were separate fields, we couldn't have that kind of dependent typing.

sophstad: No, that's a great point! At first I thought this would add unnecessary complexity to the front end, but that all-or-nothing approach makes things pretty clear. And changing the type signature to `Optional[Tuple[str, str]]` seems even better!

i80and:
> ?

Two `Target`s actually should never have the same `html_id`, at least on the same page: we explicitly go through and add a counter to any duplicates so they're unambiguous. The main thing is that because the `html_id` is on the `Target`, all of the TargetIdentifiers underneath share that element ID.

i80and:
@madelinezec I've added a comment and tweaked `test_program_option()` to make it more clear that we're testing --config refs!

PR #205, opened 2020-08-25
DOP-1359: Batch writes to the database

i80and:


snooty/main.py
```diff
@@ -190,32 +195,23 @@ def on_update(
...
-            self.client[self.db][COLL_ASSETS].replace_one(
-                {"_id": static_asset.get_checksum()},
-                {
-                    "_id": static_asset.get_checksum(),
-                    "filename": str(static_asset.fileid),
-                    "data": static_asset.data,
-                },
-                upsert=True,
+        for static_asset in page.static_assets:
+            self.pending_writes.append(
```
skerschb, line 203: nice optimization.

i80and: To be fair, if we were running this over a dialup connection, it would likely be significantly slower than the previous code! :D

PR #207, opened 2020-08-31
DOP-1277: case-sensitive target matching

i80and:


i80and:
> looks great! is this the sort of thing we don't write tests for?

No, there should be tests for this -- it's just hard to test this with our current infrastructure, so I forgot. Good catch

PR #214, opened 2020-09-08
DOP-1434: Explain ambiguous target choices

i80and:
Diagnostic message is, for example, `Ambiguous target: "std:term:something". Locations: reference/glossary, https://docs.mongodb.com/manual/reference/glossary`

### DIFF
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -1,5 +1,5 @@
 import enum
-from typing import Tuple, Union, Dict, Optional
+from typing import Tuple, Union, Dict, Optional, List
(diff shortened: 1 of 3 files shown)


skerschb (commented):
do we have a test for this change?

i80and:
> do we have a test for this change?

Yes, in `test_postprocess.py` we ensure that the candidate set is correctly populated. I'd love a less ungainly post processor test framework to make a more comprehensive test, and I have one prototyped, but it's not yet in the tree.

schmalliso (approved):
This is my new favourite log update. It is marvellous.

PR #219, opened 2020-09-21
DOP-1534: Don't lowercase labels

i80and:
Also finally introduces a more granular postprocessor testing framework, and makes running the parser outside of a git repo acceptable.


snooty/parser.py
```diff
@@ -1005,9 +1005,17 @@ def __init__(
...
+            branch = subprocess.check_output(
+                ["git", "rev-parse", "--abbrev-ref", "HEAD"],
+                cwd=root,
+                encoding="utf-8",
+                stderr=subprocess.PIPE,
+            ).strip()
+        except subprocess.CalledProcessError as err:
+            logger.info("git error getting branch name: %s", err.stderr)
+            branch = "current"
+
```
sophstad, line 1018: Just a note, would be nice to abstract this from the main parser process eventually. It feels somewhat specific to the MongoDB backend approach we are using.

snooty/test_postprocess_granular.py
```diff
@@ -0,0 +1,51 @@
+"""An alternative and more granular approach to writing postprocessing tests.
+   Eventually most postprocessor tests should probably be moved into this format."""
```
sophstad, line 2: Let's make _this_ file `test_postprocess.py` and call the old one something more curmudgeonly to encourage new work here!

snooty/parser.py
i80and: oh YES absolutely!

This is a quirk that we should rethink. But right now, I just want it to not be a *fatal* quirk 😸

snooty/test_postprocess_granular.py
i80and: Good call!! Updated!

sophstad (approved):

PR #220, opened 2020-09-22
DOP-1526: Expose language "pills" in parser

sophstad:
[DOP-1526] 
- Deprecate `tabs-pillstrip` in favor of `tabs-selector` directive, which will be used to render a dropdown selector component on the page.
  - `.. tabs-pillstrip:: platforms` only appears in mms-docs, which is not slated for migration
- Add tabs selector handler to postprocess layer
  - Provide diagnostics about tabs that don't appear in every tabset
  - Save list of tabs as page-level option (since the [designs](https://mongodb.invisionapp.com/share/NAYQ7EZY36F#/screens) include the dropdown as a third panel, it's preferable to put this data here than deep in the AST)
- **Limitation:** this approach doesn't work if `.. tabs-selector::` appears _after_ a set of tabs. I think this is okay, but let me know if you find it problematic! It can be fixed 😌
- If there's a way to make `tabs-pillstrip` (deprecated) inherit `tabs-selector` for its `name` internally, that would be great! I couldn't come up with a clean approach to handling that.
- This is somewhat related to the DOP-1450 work; it would be nice for the page option to include the tab title along with the tab ID.


snooty/diagnostics.py
```diff
@@ -331,3 +331,31 @@ def __init__(
...
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        tabs: Set[str],
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(
+            f"The following tab(s) did not appear in every tabset: {tabs}", start, end,
```
i80and, line 346: `tabset` might be a bit jargon-y -- perhaps `one or more set of tabs on this page was missing the following required tabs:`

snooty/n.py
```diff
@@ -197,7 +197,7 @@ class SubstitutionReference(InlineParent):
 class Root(Parent[Node]):
     __slots__ = ("options",)
     type = "root"
-    options: Dict[str, str]
+    options: Dict[str, Any]
```
i80and, line 200: I really don't want `Any` in the node fields: once we hit the AST, everything should be safely typed. Can we refine this type?

snooty/rstparser.py
```diff
@@ -687,6 +687,10 @@ def run(self) -> List[docutils.nodes.Node]:
 
             return [node]
 
+        # these directives should not be treated like a tabs directive
+        if self.name == "tabs-pillstrip" or self.name == "tabs-selector":
```
i80and, line 691: You can just do `if self.name in {"tabs-pillstrip", "tabs-selector"}` to make the intent slightly more clear

i80and (changes requested):

i80and (approved):
This will require a minor bump, since technically the AST schema has changed

PR #224, opened 2020-09-28
DOP-1434: Automatically resolve ambiguous targets

i80and:
This PR also copies all included ASTs upon inclusion, correcting the previous behavior.

This adds roughly 1.5s to the postprocessor, which is small but significant and repeatable for that stage taken on its own.

```
Before:
parse rst      40.20
copy           3.14
postprocessing 10.31 <-- Here
commit         0.00
serialization  3.74

After:
parse rst      38.71
copy           2.88
postprocessing 11.84 <-- Here
commit         0.00
serialization  3.54
```

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -947,7 +947,7 @@ class PageDatabase:
         postprocessor = self.postprocessor_factory()
 
         with util.PerformanceLogger.singleton().start("copy"):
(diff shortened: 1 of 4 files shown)


schmalliso (approved):
LGTM! :yay:

sophstad (approved):

PR #225, opened 2020-09-28
DOP-1523: Permit version admonitions to have no argument

i80and:
This change scares me -- admonitions are a very gnarly and fragile part of docutils. We have a lot of test cases, but likely not every single situation a writer has ever come up with.

### DIFF
--- a/snooty/rstparser.py
+++ b/snooty/rstparser.py
@@ -762,14 +762,14 @@ class BaseCodeDirective(docutils.parsers.rst.Directive):
 
 
 class BaseVersionDirective(docutils.parsers.rst.Directive):
-    """Special handling for versionadded, versionchanged, and deprecated directives.
+    """Special handling for version change directives.
 
     These directives include one required argument and an optional argument on the next line.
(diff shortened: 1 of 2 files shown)


snooty/rstparser.py
```diff
@@ -769,7 +769,7 @@ class BaseVersionDirective(docutils.parsers.rst.Directive):
     subsequent indented directives are included as children of the node.
```
sophstad, line 769: Docstring could probably stand to be changed:

> These directives include one required argument

No longer holds!

sophstad (approved):
Agree that this change is a bit 😳, but LGTM!

schmalliso (commented):
+1 to Sophie's comment about the docstring.

Does this change make including the version argument optional for versionadded and versionchanged *as well as* deprecated? Because ideally, it's only be deprecated…

i80and:
Made only `deprecated`'s argument optional, and added some more tests just to make myself feel better!

schmalliso (approved):
MAGNIFICO!

PR #227, opened 2020-09-28
DOP-1569: Warn about incorrect external link syntax

i80and:


snooty/diagnostics.py
```diff
@@ -342,3 +342,14 @@ def __init__(
...
+
+class IncorrectLinkSyntax(Diagnostic):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__("Malformed external link", start, end)
```
sophstad, line 355: It may be useful to provide an example of how external links should look here.

sophstad (approved):
One optional note!

snooty/diagnostics.py
i80and: That's actually done by the upcoming Did You Mean feature! Do you think it's important to both describe the correct syntax as well as show a corrected version?

PR #230, opened 2020-09-30
DOP-1018: Preserve source fileid through includes

i80and:
This changes the AST so the Root node contains a `fileid` field indicating the source of this document. During postprocessing, these values are tracked in a stack, and each SAX handler gets the full fileid stack history for the current page context.


snooty/eventparser.py
```diff
@@ -30,16 +58,16 @@ def get_event_listeners(self, event: str) -> Set[Callable[..., Any]]:
     def fire(
         self,
         event: str,
-        filename: FileId,
+        fileid: Union[FileId, FileIdStack],
```
sophstad, line 61: Do you see any benefit to only using a `FileIdStack` (no union) here? It seems like the page enter/exit events could work with a stack of length 1. But maybe I am missing something!

snooty/gizaparser/test_steps.py
```diff
@@ -9,13 +9,29 @@
...
+    #     category.add(path, text, steps)
+    #     assert len(parse_diagnostics) == 0
+    #     assert len(steps) == 4
+    #     return parse_diagnostics
+
+    # def add_child_file() -> List[Diagnostic]:
+    #     steps, text, parse_diagnostics = category.parse(child_path)
+    #     category.add(child_path, text, steps)
+    #     assert len(parse_diagnostics) == 0
+    #     return parse_diagnostics
```
sophstad, line 34: Remove?

snooty/n.py
```diff
@@ -45,6 +47,18 @@
 _T = TypeVar("_T")
 
 
+class FileId(PurePosixPath):
```
sophstad, line 50: Nice, this is a better home for this class.

sophstad (approved):
Nice, this looks good to me! Just left one thought.

In terms of breaking changes, I think we'll just need to ensure that the front end safely ignores the new `root` nodes. Thanks a lot for this Andrew!

snooty/gizaparser/test_steps.py
i80and: oops, yup! 😅

snooty/eventparser.py
i80and: No real strong opinion, so I just went ahead and changed the PR to use FileIdStack everywhere!

PR #231, opened 2020-10-01
DOP-1586: Support targets with escaped angle brackets

i80and:


schmalliso (approved):
My only suggestion would be to leave a comment indicating why we're doing this (i.e. docutils is crazy and adds whitespace for some reason).

i80and:
> My only suggestion would be to leave a comment indicating why we're doing this (i.e. docutils is crazy and adds whitespace for some reason).

Ah dang I forgot to do this before merging -- I'll do a followup commit.

Interestingly it's not whitespace, it's a *null byte*, which is invisible and is actually the standard string termination character in C. Its use here is a fascinating little hack.

PR #232, opened 2020-10-01
DOP-1589: Support required arguments

i80and:
Make rstobject directive arguments required. Currently, the parser crashes if no argument is passed to a callable rstobject.

### DIFF
--- a/snooty/rstparser.py
+++ b/snooty/rstparser.py
@@ -998,10 +998,24 @@ def make_docutils_directive_handler(
     name: str,
     options: Dict[str, object],
 ) -> Type[docutils.parsers.rst.Directive]:
+    optional_args = 0
+    required_args = 0
+
+    argument_type = directive.argument_type
+    if argument_type:
+        if (
+            isinstance(argument_type, specparser.DirectiveOption)
+            and argument_type.required
+        ):
+            required_args = 1
+        else:
+            optional_args = 1
+
     class DocutilsDirective(base_class):  # type: ignore
         directive_spec = directive
         has_content = bool(directive.content_type)
-        optional_arguments = 1 if directive.argument_type else 0
+        optional_arguments = optional_args
+        required_arguments = required_args
         final_argument_whitespace = True
         option_spec = options
 
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -127,7 +127,7 @@ example = """.. versionadded:: ${1:v1.0}
 [directive.deprecated]
 argument_type = {type = "string", required = false}
 content_type = "block"
-example = """.. deprecated:: ${1:v1.0}
+example = """.. deprecated:: ${1:v1.0 (Optional)}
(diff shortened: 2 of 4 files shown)


snooty/rstspec.toml
```diff
@@ -109,21 +109,23 @@ example = """.. tip:: ${1:Title}
...
-content_type = "block"
+inherit = "_baseversiondirective"
 example = """.. versionadded:: ${1:v1.0}
    ${2:Describe what was added in that version. Description is optional.}
 """
 [directive.deprecated]
-argument_type = "string"
+argument_type = {type = "string", required = false}
 content_type = "block"
 example = """.. deprecated:: ${1:v1.0}
```
schmalliso, line 130: If you were keen, you could indicate that the version is optional in this example string.

snooty/rstparser.py
```diff
@@ -998,10 +998,24 @@ def make_docutils_directive_handler(
...
+
+    argument_type = directive.argument_type
+    if argument_type:
+        if (
+            isinstance(argument_type, specparser.DirectiveOption)
+            and argument_type.required
+        ):
+            required_args = 1
+        else:
+            optional_args = 1
```
schmalliso, line 1012: This seems to make arguments optional everywhere except where they're explicitly required… Is it really only versionchanged and versionadded where arguments should be required? Don't we need an argument for `option`, `program`, and `code-block`?

schmalliso: Oh wait, this is making sure it doesn't *crash*, not that we don't complain about it. NEVER MIND, I UNDERSTAND.

snooty/test_parser.py
```diff
@@ -1705,6 +1705,24 @@ def test_callable_target() -> None:
 </root>""",
     )
 
+    # Ensure that a missing argument doesn't crash
```
schmalliso, line 1708: So do we complain in the log if someone does
```
.. method:: 

   creates an index
```

Because I feel like we should… maybe that's handled elsewhere?

schmalliso (commented):
I have a lot of questions, most of which are mostly for my edification, and one suggestion which is of substance.

snooty/test_parser.py
i80and: We will with this change -- we check that the right diagnostic is raised a bit down!

snooty/rstparser.py
i80and: So right now all arguments are optional. This change allows us to opt things into being required. `option` and `program` are rstobjects, which this change forces to be argument required.

snooty/rstspec.toml
i80and: I don't actually really understand the example string format -- could I have an example please? :D

i80and: How is this?

snooty/test_parser.py
schmalliso: aaaah excellent

schmalliso (approved):

PR #234, opened 2020-10-02
DOP-1595: Correctly inherit YAML ref names

i80and:
This also fixes multi-level inheritance, and adds explicit cycle-detection.


madelinezec (approved):
lgtm

PR #236, opened 2020-10-08
DOP-1581: extlinks should use target for label if no label set *not* the uri

schmalliso:
Jira link: https://jira.mongodb.org/browse/DOP-1581

- extlinks will use the 'target' for the label if no label is set

- handles special case for `:rfc:`, where we'd ideally like the label to be `RFC-123465` rather than just `123465`. This brings parity to sphinx's built-in rfc role and will generally behave the way writers expect it to.

It is entirely possible that there's an easier way to know what role we're parsing, but I couldn't figure it out, so this was my solution 😁

### DIFF
--- a/snooty/rstparser.py
+++ b/snooty/rstparser.py
@@ -54,6 +54,7 @@ RoleHandlerType = Callable[
 PAT_EXPLICIT_TITLE = re.compile(
     r"^(?P<label>.*?)\s*(?<!\x00)<(?P<target>.*?)>$", re.DOTALL
 )
+PAT_ROLE_NAME = re.compile(r"^:[a-zA-Z]*:")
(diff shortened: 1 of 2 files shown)


snooty/rstparser.py
```diff
@@ -373,10 +384,14 @@ def __call__(
         content: List[object] = [],
     ) -> Tuple[List[docutils.nodes.Node], List[docutils.nodes.Node]]:
         target, label = parse_explicit_title(text)
+        role = parse_role(rawtext)
```
i80and, line 387: I think this is what `typ` is, actually!

i80and (changes requested):

snooty/rstparser.py
schmalliso: WELL THAT'S SENSIBLE

PR #238, opened 2020-10-08
Bug dop 1450: supports tabs- with options

madelinezec:
thank you @sophstad for catching this right out the gate!


madelinezec:
hold on let me fix where the test file lives!

snooty/rstparser.py
```diff
@@ -1072,29 +1072,23 @@ def register_spec_with_docutils(
         )
         builder.add_directive(name, DocutilsDirective)
 
+    # copy tabs directive declaration
```
i80and, line 1075: This is referencing it, not copying it. Comments should also explain why you're doing this, if possible

snooty/rstparser.py
```diff
@@ -1072,29 +1072,23 @@ def register_spec_with_docutils(
...
+    tabs_directive = spec.directive["tabs"]
+    tabs_directive.options["tabid"] = specparser.PrimitiveType.string
```
i80and, line 1077: Is `tabid` actually an option to `tabs`?

snooty/rstparser.py
```diff
@@ -1072,29 +1072,23 @@ def register_spec_with_docutils(
...
-            example=None,
-            content_type="block",
-            argument_type=None,
-            required_context=None,
-            domain=None,
-            name=tabs_name,
-            options={"tabid": specparser.PrimitiveType.string},
-        )
+        tabs_name = "tabs-" + name
+        tabs_directive.name = tabs_name
```
i80and, line 1083: Here's where you would actually want to copy it. The way you would do this might be:
```
modified_tabs_directive = dataclasses.replace(tabs_directive, name=tabs_name)
```

using [dataclasses.replace](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace)

snooty/rstparser.py
```diff
@@ -1072,29 +1072,23 @@ def register_spec_with_docutils(
...
 
         tabs_options: Dict[str, object] = {
             option_name: spec.get_validator(option)
-            for option_name, option in directive.options.items()
+            for option_name, option in tabs_directive.options.items()
```
i80and, line 1087: then these would reference the above

i80and (changes requested):

i80and (approved):

PR #239, opened 2020-10-12
DOP-1632: Fix tab sort

sophstad:
[DOP-1632] Fix sorting of tabs by using key of tabids directly from `rstspec.toml`.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -446,7 +446,6 @@ class JSONVisitor:
         except KeyError:
             self.diagnostics.append(UnknownTabset(tabset, line))
             return
-        tabid_list: List[str] = []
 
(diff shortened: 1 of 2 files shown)


madelinezec (commented):
this looks great! interestingly there is an existing test to test ordering in `test_data/test_tabs` / [test_parser](https://github.com/mongodb/snooty-parser/blob/master/snooty/test_parser.py#L52-L62)

Just out of curiosity, could I see the rst that didn't sort correctly before approving the PR?

sophstad:
@madelinezec Definitely! The easiest way to do that would probably be to copy `test_tabs_reorder()` and run it on the `master` branch—it should fail there.

madelinezec (approved):
lgtm

PR #241, opened 2020-10-12
Dop 1601

madelinezec:
relevant ticket: https://jira.mongodb.org/browse/DOP-1601

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -28,6 +28,7 @@ from .diagnostics import (
     UnnamedPage,
     MissingTocTreeEntry,
 )
+from . import specparser
 from . import n, util
 from .page import Page
 from .target_database import TargetDatabase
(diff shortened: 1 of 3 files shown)


i80and (approved):
🚀

snooty/postprocess.py
```diff
@@ -369,7 +370,19 @@ def handle_refs(self, filename: FileId, node: n.Node) -> None:
         # Add title and link target to AST
         target_candidates = self.targets[key]
         if not target_candidates:
+            # insert title and raise diagnostic
             line = node.span[0]
+            target_dict = specparser.SPEC.rstobject
+            target_key = f"{node.domain}:{node.name}"
+            title = node.target
+            if target_key in target_dict and target_dict[target_key].prefix:
+                title = title.replace(f"{target_dict[target_key].prefix}.", "")
```
sophstad, line 379: Just wondering what the purpose of this clause is—AFAICT it serves to provide a marginally better error message? Could be worth clarifying.

madelinezec: So legacy still generated titles in the final xml even when the link you referenced was incorrect or broken (and it generated an error message to let you know). At this stage in the parser, we don't have access to the raw rst - we have access to the key in the node's target, which is a combination of the domain and the title (here the target_key would be `param.$title`. But not all target keys have domains (or `.` for that matter), so we had to find a way of abstracting the title that wasn't just `str.replace(.)`. Instead we check if the target exists in specparser.SPEC and if so, if there's a prefix defined - if both of those are true, we remove the prefix from the start of the string --> we arrive at the title!

PR #243, opened 2020-10-15
DOP-1526: Fix double tabs-pillstrip bug

sophstad:
Fixes bug where previously encountered tabsets would be overwritten if the page included multiple `tabs-pillstrip` or `tabs-selector` directives.

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -173,7 +173,9 @@ class TabsSelectorHandler:
 
             # Avoid overwriting previously seen tabsets if another tabs-pillstrip directive is encountered
             if tabset_name in self.selectors:
-                self.diagnostics[filename].append(DuplicateDirective(node.name, 0))
+                self.diagnostics[filename].append(
+                    DuplicateDirective(node.name, node.start[0])
+                )
                 return
 
             self.selectors[tabset_name] = []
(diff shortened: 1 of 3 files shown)


snooty/postprocess.py
```diff
@@ -169,6 +169,11 @@ def __call__(self, filename: FileId, node: n.Node) -> None:
             # Handle naming discrepancy between .. tabs-pillstrip:: languages and .. tabs-drivers::
             if tabset_name == "languages":
                 tabset_name = "drivers"
+
+            # Avoid overwriting previously seen tabsets if another tabs-pillstrip directive is encountered
+            if tabset_name in self.selectors:
```
i80and, line 174: Should we raise a diagnostic?

snooty/postprocess.py
```diff
@@ -172,6 +173,7 @@ def __call__(self, filename: FileId, node: n.Node) -> None:
 
             # Avoid overwriting previously seen tabsets if another tabs-pillstrip directive is encountered
             if tabset_name in self.selectors:
+                self.diagnostics[filename].append(DuplicateDirective(node.name, 0))
```
i80and, line 176: Should this be line `node.start` instead of 0?

PR #244, opened 2020-10-20
DOP-1575: Add mongo web shell as a directive

rayangler:
[[DOP-1575](https://jira.mongodb.org/browse/DOP-1575)]

This PR introduces a new `mongo-web-shell` directive. The `docs` repo currently renders the MongoDB Web Shell by having an `iframe` within a `raw` directive, which is ignored in next gen. This new directive will allow the snooty frontend to render this component wherever it is used.

### DIFF
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -468,3 +468,17 @@ class DuplicateDirective(Diagnostic):
             f"""Directive "{name}" should only appear once per page""", start, end,
         )
         self.name = name
+
+
+class InvalidMWSVersion(Diagnostic):
+    severity = Diagnostic.Level.error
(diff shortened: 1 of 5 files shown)


i80and (changes requested):
I think this can all be replaced with an enum!

Note these lines: https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml#L4-L12

This allows you to define a list of supported values for an option. So e.g.

```
[enum]
mws_version = ["3.4", "3.6", ...]

[directive."mongodb:mongo-web-shell"]
...
options.version = "mws_version"
...
```

i80and (approved):

PR #248, opened 2020-11-02
DOP-1318: add created_at field to snooty documents

schmalliso:
Adds a created_at field to snooty documents!

Evidence:
Screen Shot 2020-11-02 at 4 27 57 PM

### DIFF
--- a/snooty/main.py
+++ b/snooty/main.py
@@ -29,6 +29,7 @@ from collections import defaultdict
 from pathlib import Path, PurePath
 from typing import Any, Dict, List, Optional, Union
 from docopt import docopt
+from datetime import datetime
 
 from . import __version__, language_server
 from .parser import Project
@@ -202,6 +203,7 @@ class MongoBackend(Backend):
             "ast": page.ast.serialize(),
             "source": page.source,
             "static_assets": checksums,
+            "created_at": datetime.utcnow(),
         }
 
         if page.query_fields:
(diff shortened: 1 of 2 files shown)


snooty/main.py
```diff
@@ -202,6 +203,7 @@ def on_update(
             "ast": page.ast.serialize(),
             "source": page.source,
             "static_assets": checksums,
+            "created_at": datetime.utcnow(),
```
sophstad, line 206: The `on_update_metadata()` function performs a similar operation on the `metadata` collection (where we store things like TOCs, breadcrumbs, and other inter-site data). Although the collection is much smaller, I imagine we ought to perform a similar culling operation on it so that we expire _all_ documents associated with a build.

So basically we can add this line there, too!

sophstad (changes requested):

snooty/main.py
schmalliso: Logical! I admit I had no idea that collection existed :D Done!

sophstad (approved):
Interested in @i80and's opinion on this change to `SerializableType`, but otherwise LGTM!

i80and:
> Interested in @i80and's opinion on this change to `SerializableType`, but otherwise LGTM!

I actually suggested this :D I don't *love* this approach and am mildly concerned about this field accidentally ending up in a JSON RPC call, but we'll cross that bridge when we come to it.

PR #250, opened 2020-11-03
DOP-277: Suggest blockquote directive when unexpected indentation is encountered

sophstad:
[DOP-277] When an unexpected indentation is encountered, provide a more helpful error message by suggesting use of the `.. blockquote::` directive via `MakeCorrectionMixin`.


i80and (approved):
Rock on!! 🤘

PR #251, opened 2020-11-03
DOP-1643: Store static_assets hashes along with a reference string

i80and:


snooty/types.py
```diff
@@ -45,6 +45,9 @@ class SnootyError(Exception):
 
 @dataclass
 class StaticAsset:
+    # "key" must *exactly* match an identifier with which this asset is referred to in source text.
+    key: str
+
```
sophstad, line 50: [nit] Accidental newline?

sophstad (commented):
Wondering if you think it makes sense to remove the `filename` field from a document in the `assets` collection (i.e. [this](https://github.com/mongodb/snooty-parser/blob/master/snooty/main.py#L223)). A few reasons in favor:

- We explicitly support using multiple filenames to refer to the same image, so this field is not guaranteed to be correct
- The front end will no longer use this field when writing image files

snooty/types.py
i80and: Intentional -- there's not really a good way to clearly document instance variables, so my convention is to have a newline after each documented instance variable

i80and:
> Wondering if you think it makes sense to remove the `filename` field from documents in the `assets` collection (i.e. [this](https://github.com/mongodb/snooty-parser/blob/master/snooty/main.py#L223)). A few reasons in favor:
> 
>     * We explicitly support using multiple filenames to refer to the same image, so this field is not guaranteed to be correct
> 
>     * The front end will no longer use this field when writing image files
> 
> 
> We could also consider whether it makes sense to drop the `assets` collection entirely if this change is made.

The original intent of the `filename` field was purely debugging documentation. I'm not super involved with the ops/debugging side of things, so if losing that metadata is fine, I'm okay with removing the filename field!

sophstad (approved):

PR #252, opened 2020-11-11
DOP-1675: Don't include binary name unless requested in :option: links

i80and:
For rstobjects with type `specparser.TargetType.cmdline_option` (only std:option right now), if no label is given, just use the original target name as written for the label instead of looking it up during postprocess.


schmalliso (approved):
Ha! Well that looks great!

PR #256, opened 2020-11-17
DOP-1734: Add suffixes to heading IDs to ensure uniqueness

i80and:


sophstad (commented):
The [content source](https://github.com/schmalliso/docs-commandline-tools/blob/snooty-setup/source/release-notes/database-tools-changelog.txt) for the page linked to in the ticket heavily uses target identifiers, specifically target identifiers that _match the id generated for each heading_. I definitely think we should add some test cases for those, and I also think it may change the scope of the ticket 😅 I believe that this issue is ultimately related to DOP-1734.

sophstad (approved):
nvm^ (◕‿◕✿) LGTM!

schmalliso (approved):

PR #257, opened 2020-11-17
DOP-1566: Don't rely on docutils header IDs

i80and:
This has, in practice, two effects:

* The whole `idX` thing is banished from heading IDs
* Headings that start with an "odd" character have an ID that lead with a dash, since our `make_id()` function is quite simple and doesn't strip out leading strangeness.

Open question: should we be lower-casing the header IDs? Sphinx does as does this PR, but do we want to? This is the only ID we lowercase.


sophstad (approved):
Perfect! This seems like an ideal solution to me 👍 

Wrt lowercasing IDs: I think that I slightly prefer lowercasing them, as most characters in a URL tend to be lowercase. But I'm ultimately very neutral, so whatever is easiest!

PR #258, opened 2020-11-18
DOP-323: Handle include options for RST files

mango-db:
[DOP-323](https://jira.mongodb.org/browse/DOP-323). 
- [Docs-manual staging -- db.collection.mapReduce](https://docs-mongodbcom-staging.corp.mongodb.com/master/manual/cgoecknerwald/master/reference/method/db.collection.mapReduce
)
- [Docs-manual RST -- db.collection.mapReduce](https://github.com/mongodb/docs/blob/63a277f66118666105eb7399ad28bf47a78ba54d/source/reference/method/db.collection.mapReduce.txt)
- [Docs-manual staging - reference/mapReduce](https://docs-mongodbcom-staging.corp.mongodb.com/master/manual/cgoecknerwald/master/reference/command/mapReduce)
- [Docs-manual RST - reference/mapReduce](https://github.com/mongodb/docs/blob/63a277f66118666105eb7399ad28bf47a78ba54d/source/reference/command/mapReduce.txt)
- [Included file - includes/parameter-map-reduce.rst](https://raw.githubusercontent.com/mongodb/docs/master/source/includes/parameters-map-reduce.rst)

You can see the impact of this change by searching for "Requirements for the map function" on the page - it only appears once. Prior to this change, it appeared 4 times. Honestly, I think this included file should just be broken up into 4 files, but such is life.

Changes:
- Add `InvalidInclude` diagnostic
- Add `Comment` and `Label` nodes and appends to AST
- Add `bound_included_AST` function to bound the AST based on `start_after_text` and `end_before_text`
- Add assertions to `populate_include_nodes` function to catch failure to locate specified text and reversed bounds
- Update `rstspec.toml` to reflect actual use of `:start-after:` and `:end-before:` in both `include` and `literalinclude` directives
- Flesh out `test_include()` in `parser.py`
- Add 3 passing tests in `test_include_subset` in `test_postprocess.py` (based on comments, paragraphs, and labels)
- Add 3 failing tests in `test_include_subset_fails` in `test_postprocess.py`(based on failure to locate `:start-after:` text, failure to locate `:end-before:` text, and reversed bounds)


snooty/postprocess.py
```diff
@@ -648,7 +702,45 @@ def get_include_argument(node: n.Directive) -> str:
...
             ast = include_page.ast
             assert isinstance(ast, n.Parent)
-            node.children = [util.fast_deep_copy(ast)]
+
+            deep_copy_children: MutableSequence[n.Node] = [util.fast_deep_copy(ast)]
+
+            # TODO: Move subsgraphing implementation into parse layer, where we can
+            # ideally take subgraph of the raw RST
+            start_after_text = (
+                node.options["start-after"] if "start-after" in node.options else ""
```
sophstad, line 711: You can use Python dict's [`get()`](https://docs.python.org/3/library/stdtypes.html#dict.get) to safely attempt to retrieve this value. If it's not found, `None` will be returned, which also seems a bit cleaner to me than an empty string.
```suggestion
                node.options.get("start-after")
```

snooty/postprocess.py
```diff
@@ -648,7 +702,45 @@ def get_include_argument(node: n.Directive) -> str:
...
+            )
+            end_before_text = (
+                node.options["end-before"] if "end-before" in node.options else ""
```
sophstad, line 714: Ditto

snooty/postprocess.py
```diff
@@ -648,7 +702,45 @@ def get_include_argument(node: n.Directive) -> str:
...
+            if start_after_text or end_before_text:
+                try:
+                    # Returns a subgraph of the AST based on text bounds
+                    deep_copy_children, any_start, any_end = self.bound_included_AST(
+                        deep_copy_children, start_after_text, end_before_text
+                    )
+                    # Confirm that we found all specified text (with helpful diagnostic message if not)
+                    msg = (
+                        "If you are sure your text is within the included RST file, try using it in a comment or "
+                        "label. Search is case-sensitive."
```
sophstad, line 726: [nit] Newline ➡️ good use case for an [f-string](https://www.python.org/dev/peps/pep-0498/).

snooty/postprocess.py
```diff
@@ -615,6 +616,59 @@ def add_titles_to_label_targets(
...
+            """Helper function to determine if the given node contains specified start-after or end-before text. 
+            
+            Note: For now, we are only splicing included files based on Text and TargetIdentifier nodes. 
+            Comments have Text nodes as children; Labels have TargetIdentifiers as children. 
+            So, writers can splice on these parent directives as well."""
+            if isinstance(node, n.Text) and node.value:
+                return search_text == node.value
+            elif isinstance(node, n.TargetIdentifier):
+                if node.ids and isinstance(node.ids[0], str):
+                    return search_text in node.ids
```
sophstad, line 640: It seems like it may be clearer to search for a `Comment` or `Label` class, and then dive into their children to check for a matching string. This would avoid evaluating many `Text`/`TargetIdentifier` nodes that are just red herrings, and feels more along the lines of what a reader would expect to see here given the criteria for start-after/end-before.

snooty/postprocess.py
```diff
@@ -615,6 +616,59 @@ def add_titles_to_label_targets(
                     target.children = heading.children
             self.pending_targets = []
 
+    def bound_included_AST(
```
sophstad, line 619: I feel like this function is trying to reinvent the wheel: the purpose of the Postprocessor class is to iterate through a page's AST and perform an operation on it. Instead of writing a _new_ function to dive through the children, a new handler class could perform the work of identifying start-after/end before nodes as the event parser moves through them. Happy to chat more to flesh this out!

sophstad (changes requested):
I think that `bound_iincluded_AST()` can be simplified to really leverage the postprocessor pattern. Left a comment to that effect! Let me know if you already tried this and ran into problems.

Also! This suggests to me that we need a corresponding front end PR to ignore these nodes (and prevent warnings that they're not implemented):

> Add Comment and Label nodes and appends to AST

Great job providing useful, contextualized diagnostics! Exciting 🎉

snooty/postprocess.py
```diff
@@ -615,6 +616,59 @@ def add_titles_to_label_targets(
...
+                    node.children, start_after_text, end_before_text
+                )
+                node.children = children
+            if is_start or has_start:
+                any_start = True
+                start_index = i
+            if is_end or has_end:
+                any_end = True
+                end_index = i
+        assert (
```
i80and, line 666: The assert statement isn't safe to use in this kind of context: it is removed by the compiler in optimization mode. Assertions are to catch cases where the *developer* (you or me) has made a mistake, not where the user has provided something unexpected

mango-db: TIL! Thanks :)

snooty/postprocess.py
mango-db: Note: leaving as regular string due to lint warning: `"f-string is missing placeholders"`

snooty/postprocess.py
mango-db: To be clear, are we invoking `-O` to remove asserts at runtime? Because when I put an `assert False` in `postprocess.py`, it catches when I build. Is this a continuous integration thing? (Also if `-O` is passed somewhere, can you point me to where it is?)

i80and: We don't run with `-O`, but correct software needs to.

This also just isn't what assertions are for.

snooty/postprocess.py
```diff
@@ -123,6 +124,149 @@ def __call__(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+            argument_list = node.argument
+            assert len(argument_list) > 0
+            return argument_list[0].value
+
+        if not isinstance(node, n.Directive) or not node.name == "include":
+            return
+
+        argument = get_include_argument(node)
+        include_slug = clean_slug(argument)
+        include_fileid = self.slug_fileid_mapping.get(include_slug)
```
sophstad, line 216: Optional way to improve messy code that you didn't write: I think we can just use `FileId(include_slug)` rather than using this object as a lookup, as is done [here](https://github.com/mongodb/snooty-parser/blob/master/snooty/postprocess.py#L409).

But no need to block release if you don't want to do this!

snooty/postprocess.py
```diff
@@ -123,6 +124,149 @@ def __call__(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+            if include_fileid is None:
+                return
+
+        include_page = self.pages.get(include_fileid)
+        assert include_page is not None
+        ast = include_page.ast
+        assert isinstance(ast, n.Parent)
+        deep_copy_children: MutableSequence[n.Node] = [util.fast_deep_copy(ast)]
+
+        # TODO: Move subsgraphing implementation into parse layer, where we can
```
sophstad, line 233: [nit] Typo
```suggestion
        # TODO: Move subgraphing implementation into parse layer, where we can
```

sophstad (approved):

PR #259, opened 2020-11-19
Remove usage of unportable 'pwd' module

i80and:
On Unix the result should be essentially the same (except that `getpass.getuser()` respects environment variables), but `getpass` is portable to other platforms.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -1,10 +1,10 @@
 import collections
 import docutils.nodes
+import errno
+import getpass
 import logging
 import multiprocessing
 import os
-import errno
-import pwd
 import re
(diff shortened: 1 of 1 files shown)


mango-db (approved):
LGTM- from [getpass.getuser() docs](https://docs.python.org/3/library/getpass.html#getpass.getuser): "This function checks the environment variables `LOGNAME`, `USER`, `LNAME` and `USERNAME`, in order, and returns the value of the first one which is set to a non-empty string. If none are set, the login name from the password database is returned on systems which support the `pwd` module, otherwise, an exception is raised." Old functionality and then some!

PR #262, opened 2020-11-19
Portability Improvements

i80and:


sophstad (approved):
LGTM! I have no way to actually test this on Windows so I'm assuming you handled that 😁 and would love to see Windows as part of our Actions workflow like we discussed!

PR #264, opened 2020-11-20
DOP-1750: Fix :copyable: true handling

sophstad:
[DOP-1750] We weren't correctly handling the case where `:copyable: true` was explicitly specified. Fix that!

### DIFF
--- a/snooty/rstparser.py
+++ b/snooty/rstparser.py
@@ -744,7 +744,7 @@ class BaseTabsDirective(BaseDocutilsDirective):
 class BaseCodeDirective(docutils.parsers.rst.Directive):
     def run(self) -> List[docutils.nodes.Node]:
         source, line = self.state_machine.get_source_and_line(self.lineno)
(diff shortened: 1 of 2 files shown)


i80and (changes requested):
I don't think the test case actually tests the change: if I revert the rstparser change but keep the test change, the test still passes.

I'd actually add the following to `test_codeblock()`:
```
    # Test copyable
    page, diagnostics = parse_rst(
        parser,
        tabs_path,
        """
.. code:: js
   :copyable:

   foo

.. code-block:: python
   :copyable: false

   foo

.. code-block:: c#
   :copyable: true

   foo""",
    )
    page.finish(diagnostics)
    assert not diagnostics
    print(ast_to_testing_string(page.ast))
    check_ast_testing_string(
        page.ast,
        """
<root fileid="test.rst">
    <code lang="js" copyable="True">foo</code>
    <code lang="python">foo</code>
    <code lang="c#" copyable="True">foo</code></root>""",
    )
```

i80and (approved):

PR #269, opened 2020-12-02
Fix crash with invalid tabs, and correct diagnostic levels

i80and:


snooty/test_parser.py
```diff
@@ -87,6 +87,24 @@ def test_tabs() -> None:
...
+   .. tab::
+
+      Tab with no ID
+""",
+    )
+    page.finish(diagnostics)
+    assert [(type(d), d.severity) for d in diagnostics] == [
+        (DocUtilsParseError, Diagnostic.Level.error)
+    ]
+
```
sophstad, line 107: What's the expected output here? Is the `tab` directive omitted?

i80and: Good question. I consider the exact output of any case where an error-level diagnostic raised to be implementation-defined. If it would make your life easier in some way, I'm happy to redefine the behavior as somewhere between "tabs directive is empty" and "this tab is excluded"

sophstad: Oops, sorry for overlooking this!

Both are equally easy to handle on the front end, but it feels preferable to me to omit each tab without `tabid`—let's go with that unless you have a different opinion. Thanks Andrew!

i80and: New logic is a little gross, but it does the trick!

snooty/test_parser.py
```diff
@@ -87,6 +84,58 @@ def test_tabs() -> None:
...
+    check_ast_testing_string(
+        page.ast,
+        """
+<root fileid="test_tabs.rst">
+    <directive name="tabs" tabset="drivers">
+        <directive name="tab" tabid="python">
+            <text>Python</text>
+            <paragraph><text>Python!</text></paragraph>
+        </directive>
+        <directive name="tab" tabid="php">
```
schmalliso, line 130: Should there be a tab for JS approximately here?

i80and: Nope: I accidentally made the test confusing. The language is called nodejs; there is no `javascript` language defined in the tabset

schmalliso: OH! That actually makes a lot of sense, and I should read more carefully :D

i80and: It was a good thing to clarify! :D

sophstad (approved):

PR #270, opened 2020-12-07
DOP-1584: manpage builder

i80and:


snooty/builders/man.py
```diff
@@ -0,0 +1,389 @@
...
+        helper(self)
+        return handler.output.getvalue()
+
+
+class Formatting(Enum):
+    BOLD = auto()
+    EMPHASIS = auto()
+
+
+class XMLToManHandler:
```
mango-db, line 79: `XMLToManHandler` -> `XMLToManualHandler`? :^)

snooty/builders/man.py
```diff
@@ -0,0 +1,389 @@
...
+
+    element: ElementType
+    children: Union[str, List["ManNode"]]
+    attributes: Dict[str, str] = field(default_factory=dict)
+
+    def translate(self) -> str:
+        """Transform this node into a troff document string."""
+        handler = XMLToManHandler()
+
+        def helper(node: "ManNode") -> None:
```
mango-db, line 57: Possibly rename `translate` and/or `helper` for clarity

mango-db, line 57: Would appreciate a brief comment for `helper`

snooty/builders/test_man.py
```diff
@@ -0,0 +1,222 @@
...
+        assert "\n\n" not in troff
+
+        try:
+            text = subprocess.check_output(
+                ["groff", "-T", "utf8", "-t", "-man"], encoding="utf-8", input=troff,
+            )
+        except FileNotFoundError:
+            pytest.xfail("groff is not installed")
+
+        text = re.sub(r".\x08", "", text)
```
mango-db, line 149: Should/could this be handled by the `normalize` function?

snooty/builders/test_man.py
```diff
@@ -0,0 +1,222 @@
...
+        troff = man.render(result.pages[fileid], "mongo", "The MongoDB Shell", 1)[
+            FileId("mongo.1")
+        ]
+
+        # Empty lines are discouraged in troff source
+        assert "\n\n" not in troff
+
+        try:
+            text = subprocess.check_output(
+                ["groff", "-T", "utf8", "-t", "-man"], encoding="utf-8", input=troff,
```
mango-db, line 144: Would appreciate a comment for clarity (or ping me about this lol, I'm curious)

snooty/builders/man.py
```diff
@@ -0,0 +1,389 @@
...
+            a = self.formatting_stack[-1]
+            if a is Formatting.BOLD:
+                self.write_raw("\\fB")
+            elif a is Formatting.EMPHASIS:
+                self.write_raw("\\fI")
+        else:
+            self.write_raw("\\f1")
+
+
+class Mapper:
```
mango-db, line 199: Comment suggestion: `Maps snooty AST nodes to intermediary ManNode`

stubs/pytest.pyi
```diff
@@ -1,10 +0,0 @@
-from typing import Any, Callable, Type
```
mango-db, line 1: Could also rm `stubs/yaml/error.pyi`

snooty/n.py
```diff
@@ -334,6 +334,11 @@ class Target(Parent[Node]):
     refuri: Optional[str]
     html_id: Optional[str]
 
+    @property
+    def identifier(self) -> Tuple[str, str]:
```
mango-db, line 338: Is it plausible that `Target.identifier()` could be accidentally conflated with `TargetIdentifier`?

mango-db (approved):
## A/C
- [x] Was able to build `man` page AST after applying appropriate values in `docs/snooty.toml`
- [x] Was able to retrieve and view the man page data with a script [1] and a command (e.g. `man ./mongo.1`)
- [x] Follows the [man page formatting convention](https://www.man7.org/linux/man-pages/man7/man-pages.7.html)
- [x] Removes unnecessary typing stub `stubs/pytest.pyi`

### For my edification::
- In `./mongo.1`, I see the function `mongo <file.js>` in all bold. Is the man page convention that arguments are italicized applicable here? I'm looking at [this page](https://www.man7.org/linux/man-pages/man7/man-pages.7.html) at "Formatting conventions for manual pages describing commands"
- Can and/or should we change`test_man.py:149`'s use of `r".\x08"` to `".[\b]"`? Not sure of cross-platform implications (Incidentally, the test still pass when one matches directly on word boundaries (`".\b"`)?)
- I'm curious about the design decision behind `Project.pages()` -- is it for efficiency? On line 1307, we only access a single page at a time, so one might naively guess that there was a way to do it with a get_page_by_fileid(fileid) method
- Are there plans to implement usage of `man mongo` or `man mongo -s 1`?
- Why define type on `self.section_depth: int = 0` but not the following two booleans?
- The format stack would be applicable if we used bold and emphasis at the same time, right? Is that ever an occurrence in man page styling?


This is friggin' dope! :) 🥇 


[1]
Attached for posterity:
```
import pymongo

metadata = pymongo.MongoClient("mongodb://localhost:27017")[<access metadata>]

for filename, value in metadata.find_one()[<access files>].items():
    with open(filename, "w" if isinstance(value, str) else "wb") as f:
        f.write(value)

 ```

snooty/n.py
i80and: This is actually dead code! Removing

stubs/pytest.pyi
i80and: I'll look at this in another change :D

sophstad (approved):

PR #276, opened 2021-01-06
DOP-1874: Ignore comment nodes in troff transform

i80and:


madelinezec (commented):
I'm about ready to LGTM - but what is troff?

i80and:
> I'm about ready to LGTM - but what is troff?

troff is the typesetting format that manpages are written in. Like TeX, but 70's.

madelinezec (approved):
I'm assuming it was intentional to not have a test! so LGTM

PR #277, opened 2021-01-08
DOP-1875: Fix altering giza substitutions in inheriting nodes

i80and:
Perform substitutions when reifying rather than while inheriting, and only in the top-most `reify()` call.


sophstad (approved):
🎉

PR #279, opened 2021-01-11
DOP-1857: Allow "reusable" references

sophstad:
[DOP-1857] Allow references/links defined with only a single underscore to be reused by name elsewhere in the corpus.

### DIFF
--- a/snooty/n.py
+++ b/snooty/n.py
@@ -369,6 +369,14 @@ class Reference(InlineParent):
 
 
 @dataclass
+class NamedReference(InlineParent):
+    __slots__ = ("refname", "refuri")
+    type = "named_reference"
+    refname: str
+    refuri: str
+
+
+@dataclass
 class Role(InlineParent):
     __slots__ = ("domain", "name", "target", "flag")
     type = "role"
--- a/snooty/n.py
+++ b/snooty/n.py
@@ -339,7 +339,7 @@ class DirectiveArgument(Parent[InlineNode]):
 
 @dataclass
 class Target(Parent[Node]):
-    __slots__ = ("domain", "name", "refuri", "html_id")
+    __slots__ = ("domain", "name", "html_id")
     type = "target"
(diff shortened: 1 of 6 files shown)


snooty/n.py
```diff
@@ -368,6 +368,14 @@ class Reference(InlineParent):
     refname: str
 
 
+@dataclass
+class NamedReference(InlineParent):
```
i80and, line 372: What children could this have?

snooty/n.py
```diff
@@ -368,6 +368,14 @@ class Reference(InlineParent):
...
+    __slots__ = ("refname", "refuri")
+    type = "named_reference"
+    refname: str
+    refuri: str
```
i80and, line 376: I... think we can actually remove the `refuri` field from `Target`, now

snooty/postprocess.py
```diff
@@ -267,6 +267,36 @@ def get_include_argument(node: n.Directive) -> str:
         node.children = deep_copy_children
 
 
+class NamedReferenceHandler:
```
i80and, line 270: Could we have a docstring briefly talking about what this does?

i80and (changes requested):
A couple questions

Also, I think we need to add a method to `builders.man.TroffNodeHandler` to handle this node; e.g. https://github.com/mongodb/snooty-parser/blob/master/snooty/builders/man.py#L391-L392

snooty/n.py
sophstad: Ah duh, copypasta mistake 😅

snooty/n.py
```diff
@@ -343,7 +343,6 @@ class Target(Parent[Node]):
     type = "target"
     domain: str
     name: str
-    refuri: Optional[str]
```
i80and, line 346: We can also remove this from __slots__ :D

i80and (approved):

PR #281, opened 2021-01-19
DOP-371: Correctly represent page flags

sophstad:
[DOP-371] Fixes a small issue that's been bothering me for a while! Page flags are currently represented as empty strings rather than booleans. E.g. `:noprevnext:` becomes this:
```python
{'noprevnext': ''}
```
instead of this:
```python
{'noprevnext': True}
```

This is possible because of #220, when the page options object was modified to allow all `SerializableType` data. Will open a small corresponding front end PR when this is merged in! Thanks 😀


sophstad:
Closing in favor of a more deterministic approach!

PR #282, opened 2021-01-22
DOP-1896: Update OpenAPI directive to allow parsing through the frontend

rayangler:
[[Jira Ticket](https://jira.mongodb.org/browse/DOP-1896)]

Docs writers will now have the option to use [Redoc](https://github.com/Redocly/redoc) to parse OpenAPI content. Writers could also use urls as the argument of the `openapi` directive like so:
```
.. openapi:: https://example.com/spec.yaml
```


snooty/parser.py
```diff
@@ -591,29 +593,39 @@ def handle_directive(
...
             )
+            swagger_parse = True if "swagger-parse" in options else False
 
             try:
-                with open(filepath) as f:
-                    openapi = OpenAPI.load(f)
+                # swagger-parse leaves parsing to the Swagger React library in the frontend
+                if swagger_parse:
+                    with open(filepath) as f:
+                        spec = json.dumps(safe_load(f))
```
rayangler, line 602: Due to issues with MongoDB not allowing keys that begin with `$` to be stored, it was decided to store OpenAPI yaml file content as a JSON string instead of an object/dictionary.

snooty/rstspec.toml
```diff
@@ -624,6 +624,7 @@ example = """.. mongo-web-shell::
 
 [directive."mongodb:openapi"]
 help = """Include a reStructuredText file's contents."""
+options.swagger-parse = "flag"
```
sophstad, line 627: Given that we're not going to be dedicating much more time to supporting the Realm version of this directive, perhaps we should make the Swagger version the default and use a flag to indicate the limited rST version is in use. (This would require a small content change in 1 place for Realm). What do you think of that?

snooty/test_openapi.py
```diff
@@ -103,3 +103,20 @@ def test_openapi() -> None:
 </text><ref_role domain="std" name="label" target="post-/auth/session"></ref_role><text> API endpoints.</text></paragraph></listItem></list></directive></section><section><heading id="responses"><text>Responses</text></heading><paragraph><literal><text>200</text></literal><text>: Successfully updated.</text></paragraph></section></directive></section></directive></root>
         """,
     )
+
+    page, diagnostics = parse_rst(
```
sophstad, line 107: [nit] This may be clearer as its own test!

snooty/parser.py
```diff
@@ -591,29 +593,39 @@ def handle_directive(
             openapi_fileid, filepath = util.reroot_path(
                 FileId(argument_text), self.docpath, self.project_config.source_path
             )
+            swagger_parse = True if "swagger-parse" in options else False
```
sophstad, line 596: [nit] 😁 
```suggestion
            swagger_parse = "swagger-parse" in options
```

snooty/parser.py
```diff
@@ -591,29 +593,39 @@ def handle_directive(
...
-                with open(filepath) as f:
-                    openapi = OpenAPI.load(f)
+                # swagger-parse leaves parsing to the Swagger React library in the frontend
+                if swagger_parse:
+                    with open(filepath) as f:
+                        spec = json.dumps(safe_load(f))
+
+                        span = (line,)
+                        openapi_spec = n.OpenAPISpec(span, spec)
+                        doc.children.append(openapi_spec)
```
sophstad, line 606: I'm a little hesitant as to whether the file contents should be appended to the node's `children` property. Thus far, `children` has always represented a series of AST nodes, never a string like this. Would love to hear from @i80and!

snooty/rstspec.toml
rayangler: Good point! I could go ahead and make the change! 

I originally went with a `snooty-parse` flag but decided against it thinking that (1) it would save Realm docs the trouble of making that change and (2) it would make sense that a directive would be parsed by the parser (snooty) being used by default. But I think it makes more sense to change the flag if Realm is really the only one expected to be using this version of the directive.

snooty/parser.py
rayangler: This part was inspired by the way `literalinclude` was parsed in that it appends a `code` node as its only child, with the contents of the file added as that node's `value`.

rayangler:
Updated the PR to reflect feedback! Will update frontend PR to accommodate changes soon

snooty/parser.py
```diff
@@ -616,42 +618,66 @@ def handle_directive(
...
-            if argument_text is None:
-                self.diagnostics.append(ExpectedPathArg(name, line))
-                return doc
+            # Parsing should be done by the OpenAPI renderer on the frontend by default
+            parse_method = options.get("parse-method", "filepath")
 
-            openapi_fileid, filepath = util.reroot_path(
-                FileId(argument_text), self.docpath, self.project_config.source_path
-            )
+            if parse_method == "url":
```
sophstad, line 624: I think determining whether the argument is a URL or filepath is straightforward enough that it can be handled on the parser. IMO, not requiring an option to switch between the two is a preferable UX experience worth implementing.

rayangler: Done!

snooty/parser.py
```diff
@@ -616,7 +618,19 @@ def handle_directive(
...
 
         elif name == "openapi":
+            # Parsing should be done by the OpenAPI renderer on the frontend by default
+            snooty_parse = options.get("snooty-parse", False)
+
             if argument_text is None:
+                # Check if argument is a url instead
+                url_argument = None
+                try:
+                    url_argument = argument[0].children[0].value
```
sophstad, line 628: Interesting, I'm just wondering — how do we know this is not a filepath?

rayangler: Good point! I can change this logic to check `argument[0].refuri`, since `argument[0]` should be a `reference` if the argument is a url.

I've only ever tested this using a filepath or a url as an argument. A filepath would be assigned as `argument_text` and a url would typically be a reference with a text node with the same url as its value.

snooty/parser.py
```diff
@@ -616,7 +618,19 @@ def handle_directive(
...
+                except:
+                    pass
+                if not (url_argument is None or snooty_parse):
+                    doc.options["isUrl"] = "True"
```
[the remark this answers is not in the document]
rayangler: I added that flag so that the [frontend could determine which value to use](https://github.com/mongodb/snooty/pull/338/files#diff-1aef327520208d3233342113f027cd9a62bfbbb52295c9e9b5bb138be5aa3b3dR20-R24) (filepath or url) and what to do with said value before passing it to Redoc.

I could change the frontend logic so that it doesn't have to rely on this flag!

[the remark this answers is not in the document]
sophstad: I think it may be preferable to change the front end, mainly because we don't particularly embrace a pattern of setting options internally in the parser. Generally, I think one expect to find all of the options enumerated in the `rstspec.toml`, so we may as well not change tack!

snooty/parser.py
sophstad: Ah, thanks for the explanation. I think checking for the `refuri` probably makes sense here, good idea!

snooty/test_openapi.py
```diff
@@ -21,15 +21,15 @@ def test_openapi_with_snooty() -> None:
         path,
         """
 .. openapi:: /test_parser/openapi-admin-v3.yaml
-   :snooty-parse:
+   :uses-rst:
```
sophstad, line 24: I think this ended up backwards somehow—the Realm OpenAPI spec uses rST and the Cloud specs going forward will not. So we probably want to use the flag in the first test, sorry for the confusion Raymund!

sophstad (approved):
Nice work Raymund!! 🎉

PR #284, opened 2021-01-28
DOP-1931: Modify the toctree handling so that index.txt can add itself to the toctree with RecursionError

mango-db:
[[JIRA]](https://jira.mongodb.org/browse/DOP-1931). 

### Staging
Docs are staged with the following added to the toctree: `Overview </index>`

[[Compass]](https://docs-mongodbcom-staging.corp.mongodb.com/master/compass/cgoecknerwald/master/).
[[BI-Connector]](https://docs-mongodbcom-staging.corp.mongodb.com/master/bi-connector/cgoecknerwald/master/).

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -902,13 +902,15 @@ class Postprocessor:
 
                     toctree_node = {
                         "title": title,
-                        "slug": slug,
+                        "slug": "/" if slug == "index" else slug,
                         "children": [],
                         "options": {"drawer": slug not in self.toc_landing_pages},
                     }
 
-                    new_ast = self.pages[slug_fileid].ast
-                    self.find_toctree_nodes(slug_fileid, new_ast, toctree_node)
+                    # Don't recurse on the index page
+                    if entry.slug != "/index":
+                        new_ast = self.pages[slug_fileid].ast
+                        self.find_toctree_nodes(slug_fileid, new_ast, toctree_node)
 
                 if toctree_node:
                     node["children"].append(toctree_node)
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -845,13 +845,17 @@ class Postprocessor:
         }
         ast = self.pages[starting_fileid].ast
 
-        self.find_toctree_nodes(starting_fileid, ast, root)
+        self.find_toctree_nodes(starting_fileid, ast, root, {starting_fileid})
 
         self.toctree = root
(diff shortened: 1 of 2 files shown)


snooty/postprocess.py
```diff
@@ -902,13 +902,15 @@ def find_toctree_nodes(
...
                         "children": [],
                         "options": {"drawer": slug not in self.toc_landing_pages},
                     }
 
-                    new_ast = self.pages[slug_fileid].ast
-                    self.find_toctree_nodes(slug_fileid, new_ast, toctree_node)
+                    # Don't recurse on the index page
+                    if entry.slug != "/index":
+                        new_ast = self.pages[slug_fileid].ast
+                        self.find_toctree_nodes(slug_fileid, new_ast, toctree_node)
```
sophstad, line 913: Just thinking aloud, the idea here is to avoid recursing on a page we've already seen... could this be worth generalizing to handle that case? No problem if not!

sophstad (commented):
Just one note, and a test would be great!

snooty/postprocess.py
mango-db: Contrary to my initial thought, there's basically no performance downside to this. I parsed and measured `docs`.

#### Before change
time: `1m4.247s`
memory usage: ~`831 MB`, brief spike to `1.0GB` near the end

#### After change
time: 1m4.687s
memory usage: ~`828 MB`, brief spike to `1.0GB` near the end

mango-db:
Staging updated to reflect changes.

snooty/postprocess.py
```diff
@@ -902,20 +906,27 @@ def find_toctree_nodes(
...
-                    new_ast = self.pages[slug_fileid].ast
-                    self.find_toctree_nodes(slug_fileid, new_ast, toctree_node)
+                    # Don't recurse on the index page
+                    if slug_fileid not in visited_file_ids:
+                        new_ast = self.pages[slug_fileid].ast
+                        self.find_toctree_nodes(
+                            slug_fileid,
+                            new_ast,
+                            toctree_node,
+                            visited_file_ids | {slug_fileid},
```
sophstad, line 921: I think it may be a bit clearer to use the [`add()`](https://www.programiz.com/python-programming/methods/set/add) method instead, but your call!

sophstad (approved):
Nice, LGTM! 🚀

PR #285, opened 2021-02-01
DOP-1958: Reorder drivers tabs

sophstad:
[DOP-1958] Order drivers tabs as such:

1. MongoDB Shell
2. Compass
3. Remaining languages, alphabetically


snooty/test_postprocess.py
```diff
@@ -389,15 +389,15 @@ def test_language_selector() -> None:
         check_ast_testing_string(
             page.ast,
             """
-<root fileid="tabs-two.txt" selectors="{'drivers': {'nodejs': [{'type': 'text', 'position': {'start': {'line': 3}}, 'value': 'Node.js'}], 'c': [{'type': 'text', 'position': {'start': {'line': 3}}, 'value': 'C'}]}}">
+<root fileid="tabs-two.txt" selectors="{'drivers': {'c': [{'type': 'text', 'position': {'start': {'line': 3}}, 'value': 'C'}], 'nodejs': [{'type': 'text', 'position': {'start': {'line': 3}}, 'value': 'Node.js'}]}}">
```
schmalliso, line 392: This is just a behaviour question -- does this indicate that writers need to use all of the languages that they'll be using in tabs in the first tab declaration lest it error?

schmalliso (approved):
Just a behavioural question for my edification. LGTM!

snooty/test_postprocess.py
sophstad: Yes—specifically when using the language selector, since that really indicates a page-wide setting. If a page lacks the `tabs-selector` (or `tabs-pillstrip`) directive, then this isn't enforced.

PR #287, opened 2021-02-03
Don't crash on failed intersphinx fetch: raise diagnostic

i80and:


snooty/parser.py
```diff
@@ -1093,13 +1094,19 @@ def __init__(
     ) -> None:
         root = root.resolve(strict=True)
         self.config, config_diagnostics = ProjectConfig.open(root)
-        self.targets = TargetDatabase.load(self.config)
+        self.targets, failed_intersphinx_downloads = TargetDatabase.load(self.config)
```
sophstad, line 1097: [nit] This value is referred to as `failed_requests` elsewhere, so maybe preferable to keep it generic? But totally your call!

sophstad (approved):

PR #289, opened 2021-02-03
Fix tarball members mode

i80and:


snooty/parser.py
```diff
@@ -95,7 +95,7 @@ def bundle(
                 tar_info = tarfile.TarInfo(name=member_name)
                 tar_info.size = len(member_data)
                 tar_info.mtime = int(current_time)
-                tar_info.mode = 0x644
+                tar_info.mode = 0o644
```
sophstad, line 98: Lol 😩

sophstad (approved):

madelinezec (approved):
lgtm

snooty/parser.py
i80and: Could be worse: older convention (still used in C) is octal numbers just start with a zero :D

PR #290, opened 2021-02-08
DOP-1970: Add product landing page template

mango-db:
[DOP-1970](https://jira.mongodb.org/browse/DOP-1970).

### Added
- `kicker` directive
- `procedure`, `step` directives
- `button` directive
- PLP `rST` template
- PLP test project
- PLP tests

### Changes
- Add `:alt-style:` option to `card-group` to handle difference in card styles (done at the group level to avoid `card` style mix-n-matching)
- Change spacing and variable names `test_parser/test_tabs` and `test_tabs.rst`
- Remove trailing spaces in `test_data/test_landing/source/index.txt`


snooty/rstspec.toml
```diff
@@ -796,15 +796,17 @@ inherit = "_landing-block"
 inherit = "_landing-block"
 help = """Container to hold a group of cards"""
 options.columns = "nonnegative_integer"
+options.style = "string"
```
sophstad, line 799: I don't know that this is an appropriate use case for a string -- that prevents us from providing parser validations around its usage. Is there any reason you didn't go with a flag?

snooty/rstspec.toml
```diff
@@ -822,6 +824,36 @@ options.icon = ["path", "uri"]
...
+content_type = '_landing-block'
+
+[directive."landing:step"]
+help = 'Make a single, numbered step.'
+example = '''.. step::
+${0: Single step’s content}'''
+content_type = '_landing-block'
+
+[directive."landing:kicker"]
+help = 'Make a kicker.'
```
sophstad, line 841: This is perhaps a good place to explain more about what a kicker is 😄

snooty/test_product_landing.py
```diff
@@ -0,0 +1,171 @@
...
+    build_identifiers: BuildIdentifierSet = {}
+    with Project(
+        Path("test_data/test_product_landing"), backend, build_identifiers
+    ) as project:
+        project.build()
+
+    return backend
+
+
+def test_queryable_fields(backend: Backend) -> None:
```
sophstad, line 24: This isn't really about testing queryable fields, in devhub we extract certain directives to be queried by the front end but we're not doing anything beyond the standard here.

snooty/test_product_landing.py
```diff
@@ -0,0 +1,171 @@
+from pathlib import Path
```
sophstad, line 1: I think these tests present a risk in a sense of over-testing work that isn't actually addressed by the PLP additions -- IMO you should only be testing the new directives that are not tested elsewhere. Otherwise more work is just introduced if there's ever a deliberate change to some of these other directives.

sophstad (changes requested):

snooty/rstspec.toml
mango-db: Yes! Great question. So, I initially went with a flag (`:alt-style:`IIRC), but @schmalliso suggested that we may have more than 2 styles of cards in the future. I felt that naming, rather than flagging, could avoid future tech debt in case we want to expand card styles

sophstad: Ah, gotcha! In that case, this seems like a good use case for a new [enum](https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml#L4-L13).

sophstad (approved):

PR #296, opened 2021-03-09
DOP-1883: update URL of kotlin SDK to avoid redirect chain

schmalliso:
@nlarew Turns out we have a lovely redirect chain when crawling the Realm pages because we're linking to the wrong root URL for the Kotlin docs (from here: https://github.com/realm/realm-java/blob/1cfe291f90d1ce5c58a1ab9b44e41c225ee9f8e7/build.gradle#L238).

This will point us at the actual root folder, which will get rid of some of the jumps that we have to crawl to index the pages, apparently. If you could review, I'd be most obliged! This is only used once in the realm docs, for the TOC.


nlarew (approved):
LGTM!

PR #297, opened 2021-03-11
DOP-2045: Identify headings for 'On this page' box

sophstad:
[DOP-2045] Currently, the front end handles parsing the heading nodes that should appear in the "On this page" box. This has degraded performance of the component, as this operation is fairly expensive and really belongs in the parser. So, this PR adds that logic to the postprocess layer.

- Identify heading nodes up to _n_ specified by `:depth:`; if `:depth:` option is omitted, include all heading nodes in accordance with [docutils](https://docutils.sourceforge.io/0.4/docs/ref/rst/directives.html#table-of-contents).
- Save list of headings to `headings` field in page options

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -1,6 +1,7 @@
 import collections
 import logging
 import os.path
+import sys
 import typing
 from collections import defaultdict
 from copy import deepcopy
@@ -11,6 +12,7 @@ from typing import (
     Iterable,
     List,
     MutableSequence,
+    NamedTuple,
     Optional,
     Sequence,
     Set,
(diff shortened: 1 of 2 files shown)


rayangler (commented):
Should we change the [spec](https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml#L156) of the `contents` directive in `rstspec.toml`? Our current component doesn't really make use of certain options.

snooty/postprocess.py
```diff
@@ -301,6 +303,66 @@ def populate(self, fileid_stack: FileIdStack, node: n.Node) -> None:
         node.refuri = refuri
 
 
+class ContentsHandler:
```
i80and, line 306: A docstring would be nice to explain what this is looking for and what it does to the AST

snooty/postprocess.py
```diff
@@ -301,6 +303,66 @@ def populate(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+    class HeadingNode(NamedTuple):
```
i80and, line 307: This isn't really a node -- I'd rather have like "HeadingData" or something

snooty/postprocess.py
```diff
@@ -301,6 +303,66 @@ def populate(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+            self.current_depth += 1
+
+    def exit_section(self, fileid_stack: FileIdStack, node: n.Node) -> None:
+        if isinstance(node, n.Section):
+            self.current_depth -= 1
+
+    def __call__(self, fileid_stack: FileIdStack, node: n.Node) -> None:
+        if isinstance(node, n.Directive) and node.name == "contents":
+            self.has_contents_directive = True
+            self.contents_depth = int(node.options.get("depth", sys.maxsize))
```
i80and, line 353: Hmmm. What happens if there's multiple `contents` directives in a page?

One reasonable option might be to skip all but the first

i80and (changes requested):
This is a long time coming, sorry about this. I'm good with this -- just a couple thoughts!

snooty/postprocess.py
sophstad: This is a great point. Is there any use case for having multiple `contents` directives, or does have to do more with handling edge cases?

rayangler: It seems like there are cases of pages with 2 `contents` directive on the page ([example](https://docs.mongodb.com/manual/meta/aggregation-quick-reference/) - "On this Page" vs. "In this Section")

sophstad: Oy — the next-gen design for "On this page" definitely doesn't support something like this. Given that, I think Andrew's approach to only honor the first instance makes sense.

i80and: Hmmm, if this is something writers may reasonably expect to work based on sphinx, we should probably raise a warning to let them know.

i80and (approved):
Metal 🤘🏻

PR #299, opened 2021-03-30
DOP-2055: Add IA support

sophstad:
- Add support for `ia` and `entry` directives
  - Save their contents to page-level options, as well as to a sitewide `iatree` entry in the `metadata` document
  - If `iatree` exists, use it to generate breadcrumbs and in-order traversal of site (i.e. have it replace `toctree`)
- Add new new testing util, `check_toctree_testing_string()`, which uses an approach similar to `check_ast_testing_string()` to enable tests that are more legible than a direct dict comparison
  - Convert existing tests to use function

### DIFF
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -355,7 +355,7 @@ class MissingTocTreeEntry(Diagnostic):
         self.entry = entry
 
 
-class InvalidToctree(Diagnostic, MakeCorrectionMixin):
+class InvalidTocTree(Diagnostic, MakeCorrectionMixin):
     severity = Diagnostic.Level.error
 
     def __init__(
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -364,7 +364,7 @@ class InvalidToctree(Diagnostic, MakeCorrectionMixin):
         end: Union[None, int, Tuple[int, int]] = None,
     ) -> None:
         super().__init__(
-            "Projects with both toctree and ia directives are not supported",
+            """Projects with both "toctree" and "ia" directives are not supported""",
             start,
             end,
         )
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -355,6 +355,40 @@ class MissingTocTreeEntry(Diagnostic):
         self.entry = entry
 
 
+class InvalidToctree(Diagnostic, MakeCorrectionMixin):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(
+            "Projects with both toctree and ia directives are not supported",
+            start,
+            end,
+        )
+
+    def did_you_mean(self) -> List[str]:
+        return [".. ia::"]
+
+
+class InvalidIAEntry(Diagnostic):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        msg: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(
+            f"Invalid IA entry: {msg}",
+            start,
+            end,
+        )
+
+
 class UnknownTabset(Diagnostic):
     severity = Diagnostic.Level.error
 
@@ -535,3 +569,21 @@ class FetchError(Diagnostic):
             start,
             end,
         )
+
+
+class InvalidChild(Diagnostic, MakeCorrectionMixin):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        child: str,
+        parent: str,
+        suggestion: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(f"{child} is not a valid child of {parent}", start, end)
+        self.suggestion = suggestion
+
+    def did_you_mean(self) -> List[str]:
+        return [f".. {self.suggestion}::"]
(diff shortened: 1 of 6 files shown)


snooty/postprocess.py
```diff
@@ -542,6 +546,113 @@ def __call__(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+            if entry.name != "entry":
+                line = node.span[0]
+                self.diagnostics[fileid_stack.current].append(
+                    InvalidChild(entry.name, "ia", "entry", line)
+                )
+                continue
+
+            parsed = urllib.parse.urlparse(entry.options.get("url"))
+            if parsed.scheme:
+                url = entry.options.get("url")
```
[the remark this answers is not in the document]
sophstad: I don't think we can enforce this, sadly. `:url:` could definitely point at non-MongoDB or non-docs links (or even a docs link that would be invalid with a trailing slash, like https://docs.mongodb.com/realm-sdks/dotnet/10.0.0-beta.1/reference/Realms.Realm.html).

snooty/postprocess.py
```diff
@@ -542,6 +546,113 @@ def __call__(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+                slug = None
+            else:
+                url = None
+                slug = entry.options.get("url")
```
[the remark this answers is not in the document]
sophstad: Great point, I'll add some diagnostics/validation around this case.

snooty/postprocess.py
```diff
@@ -542,6 +546,127 @@ def __call__(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+            if len(entry.argument) > 0:
+                title = entry.argument
+            elif slug:
+                title = self.heading_handler.get_title(clean_slug(slug)) or []
+
+            snooty_name = entry.options.get("snooty-name")
+            if snooty_name and not url:
+                self.diagnostics[fileid_stack.current].append(
+                    InvalidIAEntry(
+                        "IA entry directives with snooty-name option must include url",
```
i80and, line 637: `must include the :url: option` might make this more clear

snooty/postprocess.py
```diff
@@ -637,16 +768,34 @@ def generate_metadata(self) -> n.SerializedNode:
...
+        if iatree:
+            document["iatree"] = iatree
+
         return document
 
+    def _get_page_from_slug(self, current_page: Page, slug: str) -> Optional[Page]:
+        relative, _ = util.reroot_path(
+            FileId(slug), current_page.source_path, self.project_config.source_path
+        )
+        fileid_with_ext = self.slug_fileid_mapping[relative.as_posix()]
```
i80and, line 796: Is there a possible scenario where this would throw KeyError?

i80and (changes requested):
A couple of thoughts!

snooty/postprocess.py
```diff
@@ -542,6 +546,113 @@ def __call__(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+    ) -> None:
+        self.ia: List[IAHandler.IAData] = []
+        self.diagnostics = diagnostics
+        self.heading_handler = heading_handler
+
+    def reset(self, fileid_stack: FileIdStack, page: Page) -> None:
+        self.ia = []
+
+    def __call__(self, fileid_stack: FileIdStack, node: n.Node) -> None:
+        if not isinstance(node, n.Directive) or not node.name == "ia":
```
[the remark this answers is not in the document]
sophstad: The domain on this directive looks to be an empty string, still appropriate to match you think?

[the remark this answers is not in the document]
i80and: Yep -- I think we *should* put things in domains, in general, but an empty string is still part of the canonical identifier

snooty/diagnostics.py
```diff
@@ -355,6 +355,40 @@ def __init__(
         self.entry = entry
 
 
+class InvalidToctree(Diagnostic, MakeCorrectionMixin):
```
i80and, line 358: This should be `InvalidTocTree` for consistency

snooty/postprocess.py
```diff
@@ -903,6 +1061,45 @@ def build_slug_fileid_mapping(self) -> None:
...
         self.slug_fileid_mapping = fileid_dict
 
+    def build_iatree(self) -> Dict[str, SerializableType]:
+        starting_page = self.pages.get(FileId("index.txt"))
+
+        if not starting_page:
+            return {}
+        if not isinstance(starting_page.ast, n.Root):
+            return {}
+        if not starting_page.ast.options.get("ia"):
```
i80and, line 1071: `"ia" not in starting_page.ast.options` might make the intent a little more clear

snooty/postprocess.py
```diff
@@ -903,6 +1061,45 @@ def build_slug_fileid_mapping(self) -> None:
...
+        ]
+        root: Dict[str, SerializableType] = {
+            "title": [node.serialize() for node in title],
+            "slug": "/",
+            "children": [],
+        }
+        self.iterate_ia(starting_page, root)
+        return root
+
+    def iterate_ia(self, page: Page, result: Dict[str, SerializableType]) -> None:
```
i80and, line 1085: Could we have a docstring to briefly explain the purpose and behavior of this method?

snooty/postprocess.py
```diff
@@ -1007,31 +1204,31 @@ def find_toctree_nodes(
         for child_ast in ast.children:
             self.find_toctree_nodes(fileid, child_ast, node, visited_file_ids)
 
-    def breadcrumbs(self) -> Dict[str, List[str]]:
+    def breadcrumbs(self, tree: Dict[str, SerializableType]) -> Dict[str, List[str]]:
         """Generate breadcrumbs for each page represented in the toctree"""
```
i80and, line 1208: Should this docstring be updated to `in the provided toctree`?

i80and (changes requested):
Just some style nits, then I'll lgtm!

i80and (approved):
🤘🏻

PR #301, opened 2021-04-02
Fix tests

i80and:
The old_and_monolithic tests rely on the production inventory file,
but I'm not untangling that right now.


sophstad (approved):
😎

PR #302, opened 2021-04-06
Relax docutils version constraint

felipecrs:
Refs: https://github.com/vscode-restructuredtext/snooty-parser/pull/15#issuecomment-813737514


i80and:
I'm a little nervous about this since we interface with docutils internals, but tests do pass with 0.17, so it's probably okay.

i80and:
Thank you for the PR!

PR #303, opened 2021-04-08
Dop 1878: Validation for children within ::tabs

casthewiz:
**Story**
https://jira.mongodb.org/browse/DOP-1878

**Notes**

- Adds a tab children validator for any `tabs` directive, ala the child validation logic for `tabsets`
- Alters copy of existing `tabset` warning

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -445,6 +445,9 @@ class JSONVisitor:
         ):
             self.handle_tabset(popped)
 
+        elif isinstance(popped, n.Directive) and popped.name == "tabs":
+            self.validate_tabs_children(popped)
+
         elif (
             isinstance(popped, n.Directive)
             and f"{popped.domain}:{popped.name}" == ":glossary"
@@ -916,6 +919,15 @@ class JSONVisitor:
                 InvalidTableStructure(msg, util.get_line(node) + len(node.children) - 1)
             )
             return
+
+    def validate_tabs_children(self, node: n.Directive) -> None:
+        line = node.start[0]
+        for child in node.children:
+            if (not isinstance(child, n.Directive)) or child.name != "tab":
+                self.diagnostics.append(
+                    TabMustBeDirective(str(type(child).__class__.__name__), line)
+                )
+        return
 
     def add_static_asset(self, raw_path: str, upload: bool) -> StaticAsset:
         fileid, path = util.reroot_path(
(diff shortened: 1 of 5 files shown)


snooty/parser.py
```diff
@@ -917,6 +920,15 @@ def validate_list_table(
...
             return
 
+    def validate_tabs_children(self, node: n.Directive) -> None:
+        line = node.start[0]
+        for child in node.children:
+            if (not isinstance(child, n.Directive)) or child.name != "tab":
+                self.diagnostics.append(
+                    TabMustBeDirective(str(type(child).__class__.__name__), line)
+                )
+        return
```
casthewiz, line 930: This essentially replicates the existing logic we had for validating child nodes in `handle_tabset`. This could be DRYer, but I'm not sure if there's a natural place in the repo to consolidate that logic.

snooty/test_parser.py
```diff
@@ -96,12 +97,21 @@ def test_tabs() -> None:
...
 
+        <directive name="tabs">
+            <directive name="tab" tabid="http">
+                <text>HTTP</text>
+                <paragraph><text>Tab Content</text></paragraph>
+            </directive>
+            <directive name="tip">
+                <text>This isn't a tab</text>
+            </directive>
+        </directive>
```
casthewiz, line 108: This behavior seems suspect but I was replicating what I *thought* was the existing behavior for raising a TabMustBeDirective diagnostics error.

snooty/test_postprocess.py
```diff
@@ -668,6 +668,29 @@ def test_language_selector() -> None:
             TabMustBeDirective
         ]
 
+    # Ensure we gracefully handle invalid tabs
+    with make_test(
```
sophstad, line 672: Since we don't do any validation with `TabMustBeDirective` in `postprocess.py`, this test feels a bit superfluous.

snooty/test_parser.py
sophstad: Interesting, yeah. I think the desired outcome would be to omit the invalid child entirely (though correct me if I'm wrong, @i80and). I would probably prefer to see the `tip` node omitted here.

snooty/test_postprocess.py
casthewiz: difference for this test case is using `.. tabs::` instead of `.. tabs-drivers::` 😄 

I'll differentiate the two test cases in the comments 👍

snooty/test_parser.py
i80and: I would also strongly prefer to omit the invalid node :D

There's a good argument to be made that we shouldn't be mutating node structure to that extent during AST construction, but until we have a separate pass available, it's fine.

sophstad (approved):

PR #305, opened 2021-04-20
DOP-1836: Add extra-compact card-group style

sophstad:
Adds `extra-compact` card group style to support card groups that appear on pages in docs-landing ([see InVision](https://mongodb.invisionapp.com/console/share/3EZ68OEPSKB/587712512)). For example:
image
Am very open to renaming this! I went with "extra-compact" because these cards use the same layout as `compact` but the icons are even smaller.


mango-db (approved):
I think "extra-compact" is perfectly clear. LGTM!

PR #310, opened 2021-05-06
Use the specified Python when creating a venv

i80and:
Useful for trying out alternative python implementations, e.g. `make test SYSTEM_PYTHON=$(which pyston)`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ help: ## Show this help message
 
 .venv/.EXISTS: pyproject.toml
 	-rm -r .venv snootycli.py
-	python3 -m venv .venv
+	"$(SYSTEM_PYTHON)" -m venv .venv
 	. .venv/bin/activate && \
 		python3 -m pip install --upgrade pip && \
 		python3 -m pip install flit && \


Makefile
```diff
@@ -12,7 +12,7 @@ help: ## Show this help message
 
 .venv/.EXISTS: pyproject.toml
 	-rm -r .venv snootycli.py
-	python3 -m venv .venv
+	"$(SYSTEM_PYTHON)" -m venv .venv
 	. .venv/bin/activate && \
 		python3 -m pip install --upgrade pip && \
```
rayangler, line 17: Is leaving this `python3` okay, or does it need to be the same version of python as `SYSTEM_PYTHON`?

i80and: I believe it's fine -- I don't know if this is spec'd behavior, but the venv module creates a python3 link in the venv to the appropriate binary.

Tested with CPython 3.9, Pyston 2.2.0, and PyPy 7.3.4 (PyPy requires... other changes that are out of scope of this PR though).

rayangler (approved):
LGTM!

PR #311, opened 2021-05-07
DOP-2117: Add layout option to landing Cards

mango-db:
[JIRA](https://jira.mongodb.org/browse/DOP-2117).

Currently, CardGroup layout is determined by the Card `style` option. For example, if `compact` is specified as a card style, the card layout will stack (rather than carousel) on mobile. This PR differentiates these two functions by adding a `layout` option to CardGroup.
 
Cards should "stack" vertically on mobile screens as the default option (currently used on Product Landing Pages). Writers should be able to specify 'carousel' to have cards carousel on mobile screens as a non-standard option. (See: [docs-landing](https://docs.mongodb.com) on mobile).

Expected benefits: no weird layout side effects when specifying Card `style`, ability to support new `layout`s in the future

### DIFF
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -4,9 +4,9 @@ version = 0
 [enum]
 alignment = ["left", "center", "right"]
 backlinks = ["entry", "top", "none"]
-card_type = ["small", "large"]
-card_style = ["default", "compact", "extra-compact"]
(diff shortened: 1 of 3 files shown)


rayangler (approved):
LGTM! Layout differentiation sounds good

i80and (approved):
🤘🏻

PR #313, opened 2021-05-14
DOP-2174: Add card-tag refrole

mango-db:
[DOP-2174](https://jira.mongodb.org/browse/DOP-2174).

Add card-tag role to represent the Realm landing page's [new Card tag components](https://www.figma.com/file/hPTwMb9uHOypYLckYDg072/Realm-PLP?node-id=0%3A1) (see: `Android ->`, `iOS ->`, etc). We expect the following rST:

```
.. card::
   :options: blah

   Card text.

   :card-tag:`Android <android-intro>`
```

### DIFF
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -1029,8 +1029,8 @@ type = "text"
 help = """Show a file path."""
 type = "text"
 
-[role.card-tag]
-help = """Create a tag that links to a named target."""
+[role.card-ref]
+help = """Create a styled reference that links to a named target."""
 inherit = "ref"
 
 [role.command]
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -1029,8 +1029,9 @@ type = "text"
 help = """Show a file path."""
 type = "text"
 
-[role.card-ref]
-help = """Create a styled reference that links to a named target."""
+[role."mongodb:card-ref"]
+help = """Create a reference, embedded within landing cards, with additional styling, that links to a named target."""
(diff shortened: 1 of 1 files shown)


snooty/rstspec.toml
```diff
@@ -1029,6 +1029,10 @@ type = "text"
 help = """Show a file path."""
 type = "text"
 
+[role.card-ref]
```
i80and, line 1032: Could we:
* Place this in the `mongodb` domain
* Add a help string

What directive will create the targets to be linked to?

i80and (changes requested):

snooty/rstspec.toml
schmalliso: @i80and Since she's inheriting ref, will this not work to link to refs?

i80and: @schmalliso well, it'll link to labels specifically Just making sure that's intended -- this is a new use case to me :D

schmalliso: Reasonable! Good, I've not led her completely astray then (that is the desired use-case).

mango-db: @i80and I believe there is a help string already, unless you're referring to something else?

i80and: @mango-db oh snap you're right -- sorry, I don't know how I missed that :D If we could be slightly more specific about what we mean by "styled reference" and what the intent is, that would be ideal. A help string should be enough for a writer to know when they might want to use it.

i80and (approved):
🥳

PR #315, opened 2021-06-02
DOP-2196: Remove \x00 characters from text nodes

i80and:
docutils replaces escape characters with \x00. Our two options
are to either monkey-patch docutils to prevent that (risky, brittle),
or we can undo this once we ingest docutils output.

docutils provides an unescape() function, but it does a lot of things
based on input parameters, and let's just not play that game.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -264,7 +264,9 @@ class JSONVisitor:
             if directive:
                 self.state.append(directive)
         elif isinstance(node, docutils.nodes.Text):
-            self.state.append(n.Text((line,), str(node)))
+            # docutils will inject \0000 characters into text nodes when there are escape characters
+            text = str(node).replace("\x00", "")
+            self.state.append(n.Text((line,), text))
             return
         elif isinstance(node, docutils.nodes.literal_block):
             self.diagnostics.append(
--- a/snooty/test_parser.py
+++ b/snooty/test_parser.py
@@ -2683,3 +2683,31 @@ Link to :opsmgr:`Ops Manager </page?q=true>`
 </root>
 """,
     )
+
+
+def test_escape() -> None:
+    path = ROOT_PATH.joinpath(Path("test.rst"))
+    project_config = ProjectConfig(ROOT_PATH, "")
+    parser = rstparser.Parser(project_config, JSONVisitor)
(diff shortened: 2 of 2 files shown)


schmalliso:
Just to double check, I built the server docs with this version of the parser to make sure that the issue Max Beugnet was seeing on the majority read concern reference page was handled, and it is! Solid.

mango-db:
For my understanding:

- Why are you using `|adl|`?
- Are there other escape characters that could be added to the test?
- Why is a raw string literal needed?

mango-db (approved):
LGTM! I get the correct output (appended) for [this Atlas line](https://github.com/10gen/docs-charts/blob/9b30091c154277e74481e8ad1c57aa98f981423c/source/data-sources.txt#L106):

Non-blocking nits/questions:
- Consider adding a comment providing context to the specific `|adl|` substitution test
- What other escape characters does docutils commonly replace, if any? Is there any potential for side effects of this change? 

_Appended parser output:_
```
"children": [
  {
    "type": "text",
    "position": {
      "start": {
        "line": 105
      }
    },
    "value": "Select one of the "
  },
  {
    "type": "substitution_reference",
    "position": {
      "start": {
        "line": 105
      }
    },
    "children": [
      {
        "type": "text",
        "position": {
          "start": {
            "line": 0
          }
        },
        "value": "Atlas"
      }
    ],
    "name": "service"
  },
  {
    "type": "text",
    "position": {
      "start": {
        "line": 105
      }
    },
    "value": " clusters or "
  },
  {
    "type": "substitution_reference",
    "position": {
      "start": {
        "line": 105
      }
    },
    "children": [
      {
        "type": "text",
        "position": {
          "start": {
            "line": 0
          }
        },
        "value": "Atlas Data Lake"
      }
    ],
    "name": "adl"
  },
  {
    "type": "text",
    "position": {
      "start": {
        "line": 105
      }
    },
    "value": "s."
  }
]
```

i80and:
> Non-blocking nits/questions:

>     * Consider adding a comment providing context to the specific `|adl|` substitution test
Done!

>     * What other escape characters does docutils commonly replace, if any? Is there any potential for side effects of this change?
Doesn't matter AFAIK, but I modified the test to include this!

PR #318, opened 2021-06-04
DOP-2223: Fix names for node-api roles

rayangler:
I did not realize before that the previous naming conventions of the new `node-api` roles would cause things to break. Added quotes to prevent this issue.


i80and (approved):

i80and:
We should probably have a lint for this

PR #327, opened 2021-07-02
Bump typing-extensions, use compatible semver

alexchandel:
Bump `typing-extensions`, and use "Compatible with version" semver instead of the more restrictive "Approximately equivalent to version".


i80and (approved):

i80and:
Thank you!

i80and:
@alexchandel Can you link a description of this dependency syntax? I get `Could not parse requirement: 'typing-extensions ^= 3.10.0.0'`, and it is not listed in PEP-440.

alexchandel:
Apparently caret isn't in any PEP yet.  It's an official part of [semver](https://stackoverflow.com/questions/22343224/whats-the-difference-between-tilde-and-caret-in-package-json), and supported by [pipenv, poetry](https://frostming.com/2019/01-04/pipenv-poetry/).  I cannot tell if flit does.  flit's unit tests don't even test for "~=" support.

The change I made is equivalent to `"typing-extensions >=3.10.0.0,<4.0.0.0",`.

You may want to revert and change the requirement to `"typing-extensions >=3.7.4.1,<4.0.0.0",`.

PR #329, opened 2021-07-04
Support placing YAML files in subdirectories

i80and:


i80and:
The requester for this feature is no longer at the company, and YAML in general is not something to which I want to invest this much effort in adding features.

PR #332, opened 2021-07-12
DOP-2296: Move include diagnostic logic to the postprocessor where it…

i80and:
… belongs


i80and:
This will currently cause the vscode extension to stop emitting include diagnostics. We need to turn on the postprocessor.

PR #333, opened 2021-07-12
DOP-2319 - Add new quiz directives to snooty-parser

carooobregon:
Stories/Links:
DOP-2319

Notes:
Added new directives to snooty-parser to support the new Quiz Widget


i80and:
@carooobregon could you run `make format` and commit? CI is failing

snooty/rstspec.toml
```diff
@@ -707,6 +707,26 @@ example = """.. procedure::
...
+   
+   ${2:string}"""
+argument_type = "string"
+options.quizid = {type = "string", required = true}
+
+[directive."mongodb:quizchoice"]
+content_type = "block"
+help = "A possible quiz answer and an explanation to be displayed after submission"
+example = """ .. quizchoice:: ${1:string}
+   :istrue:
```
mango-db, line 724: Nit: do we want to perhaps use kebab-case `is-true` like [so](https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml#L696), or just `true` like [so](https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml#L314)? I have no personal preferences

carooobregon: I personally like `is-true` but open to both! Changing it unless @i80and has any personal preference

i80and: I like kebab case

i80and: quiz-id should probably also get the kebab treatment if we do this

i80and (approved):

PR #336, opened 2021-08-01
DOP-2377: replacement directive

i80and:
There's three main implementation facets:
* We add a new `BlockSubstitutionReference` node that is identical to the inline `SubstitutionReference`, but it can exist in a block context.
* We *retain* replacement directives in the processed include directives, permitting them to be easily identified with that single include directive in the later substitution pass.
* We have some coersion rules: to inject inline content into a block context, we wrap it in a paragraph node. To inject block content into an inline context, we proceed only if it's a paragraph node with exclusively inline content, and if so, use that content. See `extract_inline()` for implementation.

A test in `snooty/test_postprocess_old_and_monolithic.py` is changed: the frontend shouldn't care, but invalid `SubstitutionReference` nodes now get emptied.

Behavior of how we report SubstitutionReference errors is slightly different: we now report the file that the substitution reference is in, rather than the page that ultimately includes it. This way, the line number we show isn't a complete lie. However, it behooves us to in the future show the full inclusion hierarchy for some diagnostics to help users figure out what's going on.


snooty/postprocess.py
```diff
@@ -812,67 +850,141 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
-                    self.unreplaced_nodes.append((node, line))
-
-                if self.seen_definitions is not None:
-                    self.seen_definitions.add(node.name)
-        except KeyError:
-            # If node does not contain "name" field, it is a duplicate substitution definition.
-            # An error has already been thrown for this on parse, so pass.
-            pass
+        if isinstance(node, n.Directive):
+            if node.name != "include":
```
schmalliso, line 854: Will this be updated when the sharedinclude directive gets added?

snooty/rstspec.toml
```diff
@@ -202,6 +203,14 @@ example = """..literalinclude:: ${1:/path/to/file.js}
    :lines:}
 """
 
+[directive.replacement]
+help = """Specify replacements for variables in a shared content item."""
+argument_type = "str"
+content_type = "block"
+example = """.. replacement:: ${1:variable name}
+   ${2:Content with which to replace variable}
```
schmalliso, line 211: There should be an extra newline above this :)

schmalliso (commented):
LGTM % the snippet needing an extra newline.

snooty/postprocess.py
i80and: Yepyep!

PR #340, opened 2021-08-25
Broaden networkx dependency

alexchandel:
Broaden `networkx` dependency to all compatible releases, without using flit's unsupported "Compatible with version" semver operator (#327).


alexchandel:
@i80and Good to merge?  FYI this is the syntax that flit's maintainers recommend.  (They think it's better than `~=`.)

i80and (approved):

i80and:
Thank you for the context!

PR #342, opened 2021-08-27
DOP-2424: Fix substitution lifting

i80and:
The bug observed was caused by us popping the parent Paragraph node, but continuing to call `handle_departure()` for it causing a double state-pop.

Adding an assertion in list construction also caught a somewhat obscure issue where a `snooty_diagnostic` docutils node would also cause an undesired state pop.

There were also escaping issues in `ast_to_testing_string()` causing a test failure. It would be nice to use a proper XML generator rather than... this.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -78,6 +78,16 @@ NO_CHILDREN = (n.SubstitutionReference,)
 logger = logging.getLogger(__name__)
 
 
+def eligable_for_paragraph_to_block_substitution(node: docutils.nodes.Node) -> bool:
+    """Test if a docutils node should emit a BlockSubstitutionReference *instead* of a normal
+    Paragraph."""
+    return (
+        isinstance(node, docutils.nodes.paragraph)
+        and len(node.children) == 1
+        and isinstance(node.children[0], docutils.nodes.substitution_reference)
+    )
+
+
 def bundle(
     filename: PurePath, members: Iterable[Tuple[str, Union[str, bytes]]]
 ) -> bytes:
@@ -340,6 +350,9 @@ class JSONVisitor:
                 )
             )
         elif isinstance(node, docutils.nodes.list_item):
+            assert isinstance(
+                self.state[-1], n.ListNode
+            ), "Attempting to place a list item in a non-list context"
             self.state.append(n.ListNodeItem((line,), []))
         elif isinstance(node, docutils.nodes.title):
             # Attach an anchor ID to this section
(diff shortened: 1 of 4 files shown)


snooty/parser.py
```diff
@@ -419,7 +430,7 @@ def dispatch_visit(self, node: docutils.nodes.Node) -> None:
             raise docutils.nodes.SkipNode()
         elif isinstance(node, rstparser.snooty_diagnostic):
             self.diagnostics.append(node["diagnostic"])
-            return
+            raise docutils.nodes.SkipNode()
```
rayangler, line 433: Just for my understanding: are we using `SkipNode` instead of `return` here because of the way diagnostics nodes would cause wonky behavior with removing lists but not list items?

rayangler (commented):
Just a couple of questions but the changes look good overall! The problematic [page](https://docs-mongodbcom-staging.corp.mongodb.com/snooty-test-fix-substitutions/docs/raymundrodriguez/master/core/causal-consistency-read-write-concerns/) from before seems to render fine on the frontend as well.

I did run in to the following error when parsing the `docs` repo locally:
```
ERROR:snooty.builders.man:INTERNAL: Unknown node type: BlockSubstitutionReference
```
The error shouldn't be caused by any of the changes introduced in this PR but just wanted to let you know! Not a blocker for this particular PR

snooty/parser.py
```diff
@@ -78,6 +78,16 @@
 logger = logging.getLogger(__name__)
 
 
+def eligable_for_paragraph_to_block_substitution(node: docutils.nodes.Node) -> bool:
```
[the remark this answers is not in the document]
i80and: jeez that's embarrassing :D

snooty/parser.py
i80and: So the way the recursion works is that `dispatch_visit()` creates a node to append to the state stack, and `dispatch_departure()` then pulls that node off and adds it to its parent's `children` list. Just returning here doesn't create a node, but still calls `dispatch_departure()` to pop the most recent node. SkipNode prevents calling `dispatch_departure()`.

This whole class is a hot mess, but rewriting it is out of scope right now :D

i80and:
Good catch on the manpage builder!

snooty/parser.py
rayangler: Thank you for this explanation!

rayangler (approved):
🚀

PR #343, opened 2021-08-31
DOP-2415: support entering symlinks

i80and:
It is useful to have a content directory contain symlinks. Currently, these are not followed, as we just use the default `os.walk()` behavior. This makes two changes:

* Redoes how `util.get_files()` is tested to make the test less brittle, and to contain symlinks & loops & fun things.
* `util.get_files()` will now enter symlinks, but will only enter any given directory (keyed via `Path.resolve()`) once


snooty/util.py
```diff
@@ -60,10 +61,17 @@ def reroot_path(
...
     """Recursively iterate over files underneath the given root, yielding
-    only filenames with the given extensions."""
-    for base, dirs, files in os.walk(root):
+    only filenames with the given extensions. Symlinks are followed, but
+    any given concrete directory is only scanned once."""
+    seen: Set[Path] = set()
+
+    for base, dirs, files in os.walk(root, followlinks=True):
+        dirs_set = set(Path(d).resolve() for d in dirs)
+        dirs[:] = [d.name for d in (dirs_set - seen)]
```
rayangler, line 72: What is this new value of `dirs` being used for?

i80and: Excellent question! This is an oddball eccentricity of the `os.walk()` interface: the second member of the return tuple, `dirs`, lists the subdirectories that will be explored next. But if you modify this list, you get to filter which subdirectories you actually want to explore!

Note the comments on the `topdown` argument in https://docs.python.org/3/library/os.html#os.walk

rayangler: This makes a lot more sense now. Thank you!

rayangler (approved):
LGTM

PR #344, opened 2021-09-02
DOP-2430: Harden getfiles() to avoid crawling above snooty.toml

i80and:
Symlinks exiting the project root (i.e. going above snooty.toml in the resolved filesystem hierarchy) are no longer followed


rayangler (approved):

schmalliso (approved):
Approach looks sensible; I've not read through the code thoroughly otherwise though.

PR #347, opened 2021-09-22
DOP-2443: Implement Chapters on the postprocessor

rayangler:
DOP-2443

This PR adds the new `chapters`, `chapter`, and `guide` directive to the parser. A new handler has been added to the postprocessor to take care of all guides and chapters to construct their metadata. Metadata for individual guides that include their time and descriptions will be added in a separate PR.

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -820,9 +820,6 @@ class GuidesHandler(Handler):
 
             handleChapters(node)
 
-            print("handled?")
-            print(self.chapters)
-
 
 class IAHandler(Handler):
     """Identify IA directive on a page and save a list of its entries as a page-level option."""
@@ -1381,8 +1378,6 @@ class Postprocessor:
         chapters = context[GuidesHandler].chapters
         if chapters:
             document["chapters"] = chapters
-            print("document_chapters")
-            print(document["chapters"])
 
         return document
 
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -6,6 +6,7 @@ import typing
 import urllib.parse
 from collections import defaultdict
 from copy import deepcopy
+from dataclasses import asdict, dataclass
 from typing import (
     Any,
     Callable,
@@ -716,9 +717,15 @@ class BannerHandler(Handler):
 class GuidesHandler(Handler):
     """Constructs a dictionary of chapters and their data and returns metadata on individual guides."""
 
+    @dataclass
+    class ChapterData:
+        chapter_number: int
+        description: Optional[str]
+        guides: List[str]
+
     def __init__(self, context: Context) -> None:
         super().__init__(context)
-        self.chapters: Dict[str, n.SerializedNode] = {}
+        self.chapters: Dict[str, GuidesHandler.ChapterData] = {}
 
     def __handle_chapter(self, chapter: n.Directive, current_file: FileId) -> None:
         """Saves a chapter's data into the handler's dictionary of chapters"""
@@ -752,10 +759,7 @@ class GuidesHandler(Handler):
                 MissingChild("chapter", "guide", line)
             )
 
-        # DocUtilsParseError will be appended if there is no description
-        description = chapter.options.get("description")
         title_argument = chapter.argument
-
         if not title_argument:
             self.context.diagnostics[current_file].append(
                 InvalidChapter("Title argument is empty.", line)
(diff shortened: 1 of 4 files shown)


snooty/postprocess.py
```diff
@@ -710,6 +713,120 @@ def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
...
 
+class GuidesHandler(Handler):
+    """Constructs a dictionary of chapters and their data and returns metadata on individual guides."""
+
+    def __init__(self, context: Context) -> None:
+        super().__init__(context)
+        self.chapters: Dict[str, n.SerializedNode] = {}
+
+    def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
+        def handleChapter(chapter: n.Directive) -> None:
```
i80and, line 724: Python uses snake_case, not camelCase.

It's also better to just have these be methods of GuidesHandler: otherwise the compiler has to issue far more instructions to create the closures at runtime.

snooty/postprocess.py
```diff
@@ -710,6 +713,120 @@ def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
...
+            """Saves a chapter's data into the handler's dictionary of chapters"""
+
+            guides: List[str] = []
+
+            for child in chapter.get_child_of_type(n.Directive):
+                if not child.name == "guide":
```
i80and, line 730: `!=` instead of `not ==` would be more idiomatic

snooty/postprocess.py
```diff
@@ -710,6 +713,120 @@ def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
...
+
+            line = chapter.span[0]
+
+            # A chapter should always have at least one guide
+            if len(guides) < 1:
+                self.context.diagnostics[fileid_stack.current].append(
+                    MissingChild("chapter", "guide", line)
+                )
+
+            # DocUtilsParseError will be appeneded if there is no description
```
i80and, line 756: spelling: appended

snooty/postprocess.py
```diff
@@ -710,6 +713,120 @@ def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
                 )
 
 
+class GuidesHandler(Handler):
+    """Constructs a dictionary of chapters and their data and returns metadata on individual guides."""
+
+    def __init__(self, context: Context) -> None:
+        super().__init__(context)
+        self.chapters: Dict[str, n.SerializedNode] = {}
```
i80and, line 721: I would *so* much rather have this be a typed object. We overuse the `Dict[str, n.SerializedNode]` idiom as it is. Could you define a dataclass, or better NamedTuple, and turn it into a dictionary only when actually saving it into the metadata collection?

snooty/postprocess.py
```diff
@@ -710,6 +713,120 @@ def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
...
+                    )
+                    continue
+
+                guide_slug = clean_slug(guide_argument[0].get_text())
+                guides.append(guide_slug)
+
+            line = chapter.span[0]
+
+            # A chapter should always have at least one guide
+            if len(guides) < 1:
```
i80and, line 751: `if not guides` would be more idiomatic

snooty/postprocess.py
```diff
@@ -1173,6 +1290,7 @@ class Postprocessor:
             TabsSelectorHandler,
             ContentsHandler,
             BannerHandler,
+            GuidesHandler,
```
i80and, line 1293: It's a little odd, I know, but Handlers are run on every page. If we only want to hit index.txt, we're doing a *lot* of extra work.

We can still use GuidesHandler, but instead of adding it to PASSES, could we run just this handler on `context.pages[FileId("index.txt")]` in `generate_metadata()`?

i80and (changes requested):

snooty/postprocess.py
i80and: Mm, incidentally the `dataclasses.asdict()` function automatically (shallowly) converts a dataclass to a dictionary

rayangler: Updated it to use dataclass and `asdict()`. Let me know if you'd prefer me to make it into a `NamedTuple` instead!

i80and: Dataclass is great! Thanks!

snooty/postprocess.py
rayangler: I made `GuidesHandler` with the intention of handling a few guides-related directives, including chapters, which should only be found on `index.txt`. However, I wanted the handler to also look at other directives found on actual guides pages for additional/reusable metadata, including a guide's `time` and its `short-description`, which would be used in a couple of different places. 

Would it be better to split the logic into two separate handlers for this case (one for chapters only, and another for guide-related data)?

i80and: That being the case, this is probably okay for now :+1:

snooty/postprocess.py
```diff
@@ -710,6 +714,128 @@ def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
...
+                )
+                continue
+
+            guide_slug = clean_slug(guide_argument[0].get_text())
+            guides.append(guide_slug)
+
+        line = chapter.span[0]
+
+        # A chapter should always have at least one guide
+        if not len(guides):
```
i80and, line 757: `not guides` is both more idiomatic and, less importantly, faster to evaluate

snooty/postprocess.py
```diff
@@ -710,6 +714,128 @@ def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
...
+            self.context.diagnostics[current_file].append(
+                MissingChild("chapter", "guide", line)
+            )
```
i80and, line 760: Should we return here, or is there useful work yet to be done?

i80and (changes requested):

i80and (approved):

PR #352, opened 2021-10-31
Rewrite makefile as a python script

i80and:


i80and:
Windows support would be nice, but at this point the PR should probably be rebuilt from scratch.

PR #353, opened 2021-11-03
DOP-2556: add roles required for k8s extlinks

schmalliso:


snooty/rstspec.toml
```diff
@@ -1434,6 +1434,27 @@ type = {link = "https://mongodb.github.io/node-mongodb-native/3.6/api/%s"}
...
+[role."osi"]
+type = {link = "https://opensource.org%s"}
+
+[role."osil"]
+type = {link = "https://opensource.org/licenses/%s"}
+
+[role."tldrl"]
+type = {link = "https://www.tldrlegal.com/l/%s"}
+
+[role."q-mdb"]
```
i80and, line 1452: Are these existing roles in legacy, or could we, say, call these e.g. `quay-mongodb`? I'd broadly much rather roles not be too acryonym'd

schmalliso: These exist currently in the k8s operator source; if we decide to change it we'll need to update those as well. 🤷🏻‍♀️

PR #355, opened 2021-11-09
DOP-2562: support lineno-start option

schmalliso:
Front-end support will be handled in separate PR, obviously :)


i80and (approved):

schmalliso:
There's no reason _not_ to merge, even though it doesn't do anything at this point, IMO. So I've merged.

PR #358, opened 2021-11-16
DOP-2546: Activate postprocessor in language_server

i80and:
This is a big PR, but is broken down into pretty incremental commits that probably make sense to be reviewed in sequence instead of the whole thing at once

Current build artifacts: https://github.com/mongodb/snooty-parser/suites/4441832356/artifacts/117748017


schmalliso (approved):
Looks good to I!

PR #360, opened 2021-11-19
DOP-2533: Support using Realm to fetch OpenAPI specs for the frontend

rayangler:
[Frontend Staging Link](https://docs-mongodbcom-staging.corp.mongodb.com/test-openapi/cloud-docs/raymundrodriguez/DOP-2533/reference/api/openapi/)
[Associated frontend PR](https://github.com/mongodb/snooty/pull/512) - to be merged together with this PR.

The purpose of this PR is to allow the `openapi` directive to accept an argument that acts as a key for the frontend to call a Realm function with. 

Additionally, using the `openapi` directive with a url as an argument has been updated such that the contents of the remote file will be saved as a JSON string. This is to prevent the frontend from running into CORS errors from urls that don't have the right access configuration.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -654,6 +654,8 @@ class JSONVisitor:
         elif name == "openapi":
             # Parsing should be done by the OpenAPI renderer on the frontend by default
             uses_rst = options.get("uses-rst", False)
+            # The frontend will grab the file contents from Realm, using the argument as its key
+            uses_realm = options.get("uses-realm", False)
 
             if argument_text is None:
                 # Check if argument is a url instead
@@ -662,8 +664,11 @@ class JSONVisitor:
                     url_argument = argument[0].refuri
                 except:
                     pass
-                if url_argument is None or uses_rst:
+                if (url_argument is None and not uses_realm) or uses_rst:
                     self.diagnostics.append(ExpectedPathArg(name, line))
+                return doc
+
+            if uses_realm:
                 return doc
 
             openapi_fileid, filepath = util.reroot_path(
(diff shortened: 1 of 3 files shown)


snooty/parser.py
```diff
@@ -654,16 +654,33 @@ def handle_directive(
...
+                if uses_realm or uses_rst:
+                    self.diagnostics.append(ExpectedPathArg(name, line))
+                    return doc
+
                 # Check if argument is a url instead
                 url_argument = None
+                spec = None
                 try:
                     url_argument = argument[0].refuri
+                    with urllib.request.urlopen(url_argument) as url:
```
i80and, line 670: Two issues with this:

1) urllib is actually dispreferred despite being in the stdlib: if you need to make a direct HTTP request, better to use, well, `requests` (which uses `urllib3` behind the scenes)
2) This request will be made every time the page is parsed, which will happen as the writer types. This is really problematic! It would be nice to have a proper mechanism for dealing with this kind of thing, but for now, probably better to use our `util.HTTPCache` class to at least make sure that the language server doesn't have to keep re-fetching the file every ~100ms. See: https://github.com/mongodb/snooty-parser/blob/master/snooty/intersphinx.py#L149

i80and (changes requested):

snooty/parser.py
```diff
@@ -10,7 +10,7 @@
 import subprocess
 import threading
 import time
-import urllib.parse
```
i80and, line 13: this change can be reverted

i80and (changes requested):
one little change, then g2g!

i80and (approved):

PR #363, opened 2021-12-01
Improve util.HTTPCache

i80and:
* Add a --no-cache command line option that can be used to disable caching entirely
* Don't actually send an HTTP request if the file was last updated in the past hour, as the comment states


rayangler (approved):
LGTM (as far as I can tell)

PR #366, opened 2021-12-15
DOP-2873: Add option to send output documents to a manifest file

i80and:
Frontend PR: https://github.com/mongodb/snooty/pull/620

### Staging

https://docs-mongodbcom-integration.corp.mongodb.com/master/docs/heli/DOP-2873/


snooty/main.py
```diff
@@ -322,8 +402,14 @@ def main() -> None:
...
+        if output_path:
+            os.unlink(output_path)
+        raise
     finally:
-        if connection:
+        if isinstance(backend, ZipBackend):
+            backend.close()
+        elif connection:
             print("Closing connection...")
             connection.close()
```
casthewiz, line 414: It may be worth adding `.close` on the `MongoBackend` so we can have uniform responsibility of `close` across `Backend` derived classes, and then we can just call `backend.close` without needing to differentiate

something like 
```
def close(self) -> None:
     if(self.client):
         print("Closing connection...")
         self.client.close()
```
?

snooty/main.py
```diff
@@ -1,7 +1,7 @@
 """Snooty.
 
 Usage:
-  snooty build [--no-caching] <source-path> [<mongodb-url>] [options]
+  snooty build [--no-caching] <source-path> [<mongodb-url>] [--output=<path>] [options]
```
rayangler, line 4: Should we also include `--output=<path>` as an option in the documentation?

snooty/main.py
```diff
@@ -171,93 +203,132 @@ def _config_db(self) -> str:
...
+        field: Dict[str, SerializableType],
     ) -> None:
-        super().on_update(prefix, build_identifiers, page_id, page)
-
-        uploadable_assets = [
-            asset for asset in page.static_assets if asset.can_upload()
-        ]
-
-        fully_qualified_pageid = "/".join(prefix + [page_id.without_known_suffix])
+        property_name = "/".join(prefix)
```
casthewiz, line 212: I would change this to `property_name_with_prefix` or something similarly contextual - at a glance on line 216 I assumed we removed prefixing the `page_id` and just had the property name ala `docs` or `drivers` until I double checked this line 😅

casthewiz (approved):
One small suggestion on class responsibilities and one variable name nitpick but otherwise looks 💯 !

snooty/main.py
```diff
@@ -299,10 +370,19 @@ def main() -> None:
...
+    output_path = args["--output"]
+
+    connection: Optional[pymongo.MongoClient] = None
+
+    if url:
+        connection = pymongo.MongoClient(url, password=getpass.getpass())
+        backend: Backend = MongoBackend(connection)
+    elif output_path:
+        zf = zipfile.ZipFile(output_path, mode="w")
+        backend = ZipBackend(zf)
```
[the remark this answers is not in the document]
i80and: Right -- I want to do away with the Atlas functionality entirely down the road, since I think having a pymongo dependency is a little weird for the parser.

Do you see a reason we'd want to use multiple backends in one build?

snooty/main.py
i80and: Excellent idea, thank you!

snooty/main.py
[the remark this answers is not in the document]
rayangler: Yup, makes sense!

I don't think it would be mandatory to have both, but there have been times where being able to see the AST in Atlas has been useful for debugging, especially for builds on the autobuilder. Straying away from this isn't the end of the world though since we can just run the parser locally to see output as needed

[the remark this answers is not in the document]
casthewiz: I think that in the ideal world, the autobuilder or whatever CI/CD process is utilized can bear the responsibility of uploading build artifacts into Atlas - which is a much cleaner abstraction anyways, since then it can hold responsibility over any tangential upsert logic and the parser can focus on parsing 😄

PR #367, opened 2021-12-20
(DOP-2320) Add tocicon page-level option to toctree_node

gjchong25:


snooty/postprocess.py
```diff
@@ -1683,11 +1683,27 @@ def find_toctree_nodes(
                             else None
                         )
 
+                    isSyncPillOn = False
```
i80and, line 1686: Python convention is snake_case

snooty/postprocess.py
```diff
@@ -1683,11 +1683,27 @@ def find_toctree_nodes(
...
+                                isSyncPillOn = True
+
                     toctree_node = {
                         "title": title,
                         "slug": "/" if slug == "index" else slug,
                         "children": [],
-                        "options": {"drawer": slug not in toc_landing_pages},
+                        "options": {
+                            "drawer": slug not in toc_landing_pages,
+                            "tocicon": isSyncPillOn,
```
i80and, line 1705: this should really be a string, not a boolean. see above

snooty/postprocess.py
```diff
@@ -1683,11 +1683,27 @@ def find_toctree_nodes(
                             else None
                         )
 
+                    isSyncPillOn = False
+
+                    # Check if tocicon is a page level option
+                    if context.pages[FileId(slug_fileid)].ast.options:
+                        if "tocicon" in context.pages[FileId(slug_fileid)].ast.options:
+                            if (
```
i80and, line 1691: I'd say we should just pass in the string, rather than checking it against "sync"

snooty/test_postprocess.py
```diff
@@ -478,11 +478,11 @@ def test_toctree_self_add() -> None:
             """
 <toctree slug="/">
     <title><text>untitled</text></title>
-    <toctree slug="page1" drawer="True" />
-    <toctree slug="/" drawer="True">
+    <toctree slug="page1" drawer="True" tocicon="False" />
```
i80and, line 481: If there's no tocicon set, we shouldn't have this entry in the options I think. Additionally, we should have a test *with* a tocicon. lmk if you need help with that!

i80and (changes requested):

i80and (approved):

rayangler (approved):

PR #369, opened 2021-12-21
DOP-2638: Handle escaped double-quotes

i80and:


casthewiz (approved):
Might be worth changing the chained `.replace` items to a dictionary structure and doing a  `for k, v in dict.items(): text.replace(k,v)` if this gets longer in the future, but that's just a nitpick 😄 

LGTM! 🚢

PR #375, opened 2022-01-19
(DOP-2651) Implement input, output, and io-code-block directives

gjchong25:


snooty/parser.py
```diff
@@ -965,6 +982,70 @@ def validate_tabs_children(self, node: n.Directive) -> None:
...
         return
 
+    def validate_io_code_block_children(self, node: n.Directive) -> None:
+        # new_children should contain input and output directives
+        new_children: List[n.Node] = []
+        line = node.start[0]
+        expected_children = {"input", "output"}
+
+        for child in node.children:
+            if isinstance(child, n.Directive):
```
i80and, line 992: You can get rid of this level of indentation just by having `if not isinstance(child, n.Directive):`, handling that case, and `continue`ing

snooty/parser.py
```diff
@@ -965,6 +982,70 @@ def validate_tabs_children(self, node: n.Directive) -> None:
...
+                                if "copyable" in node.options
+                                and node.options["copyable"]
+                                else False
+                            )
+                            new_grandchildren.append(grandchild)
+                        child.children = new_grandchildren
+                        new_children.append(child)
+                else:
+                    # either duplicate input/output or invalid child is provided
+                    msg = f"already contains 1 {child.name} directive"
```
i80and, line 1020: probably `an` rather than `1`.

snooty/parser.py
```diff
@@ -700,138 +705,150 @@ def create_page() -> Tuple[Page, EmbeddedRstParser]:
...
                     )
-                )
+                    return doc
 
-            # If we failed to locate end_before text, default to the end-of-file
-            if end_before == -1:
-                end_before = len(lines)
+                lines = text.split("\n")
+                # Capture the original file-length before splicing it
+                len_file = len(lines)
```
i80and, line 734: this can be defined closer to where it's used down below

snooty/parser.py
```diff
@@ -700,138 +705,150 @@ def create_page() -> Tuple[Page, EmbeddedRstParser]:
...
-                text = filepath.read_text(encoding="utf-8")
-            except OSError as err:
-                self.diagnostics.append(
-                    CannotOpenFile(argument_text, err.strerror, line)
+        elif name == "literalinclude" or name == "input" or name == "output":
+            if name == "literalinclude":
+                if argument_text is None:
+                    self.diagnostics.append(ExpectedPathArg(name, line))
+                    return doc
+            if argument_text is not None:
```
i80and, line 713: If I'm scanning this correctly, there's no else clause here, correct? Could we do e.g. `if argument_text is None: return doc` and dedent the `if` body?

snooty/rstspec.toml
```diff
@@ -463,6 +463,43 @@ example = """.. code:: ${1:language}
...
 """
 
+[directive.io-code-block]
+help = """A block of code and its corresponding output to show the user."""
+argument_type = "string"
+content_type = "block"
+options.caption = "string"
+options.copyable = "boolean"
+options.class = "string"
+options.source = "uri"
```
i80and, line 473: Is `source` to be implemented in a future PR?

i80and, line 473: Is `source` to be implemented in a future PR?

i80and (changes requested):
Good stuff!

snooty/rstspec.toml
```diff
@@ -463,6 +463,43 @@ example = """.. code:: ${1:language}
...
+
+[directive.output]
+help = """The code output."""
+inherit = "input"
+example = """.. io-code-block:: ${1:language}
+   .. output:: ${0:code output or </path/to/file>}
+      :emphasize-lines: (string)
+      :linenos: (flag)
+"""
+
```
schmalliso, line 502: I'm not sure why this didn't occur to me previously - I'm sorry. Is there a context in which we'd want the input to be one language and the output another?

/cc @terakilobyte, who is particularly enthused about this feature to weigh in

snooty/rstspec.toml
gjchong25: Yep, once the LG changes for custom action buttons for the Code component are in prod, the frontend for this can then be implemented.

gjchong25: Resolving as duplicate.

snooty/parser.py
gjchong25: Yep!

snooty/rstspec.toml
gjchong25: I assumed that the output would just be plaintext, but if we wanted writers to be able to specify a language for the output codeblock it shouldn't be hard.

snooty/parser.py
gjchong25: Actually since lines gets changed within the next if block, and len_file depends on the original file-length before it gets altered, I think it needs to stay there!

snooty/rstspec.toml
terakilobyte: I envision output to mainly be either plaintext or json, but it could be other formats. A Query API to SQL demonstration, or if MongoDB ever creates generators, or if we needed to go into deeper detail on some builders the drivers expose.

gjchong25: So would we like the input/output directives to be able to specify a language instead of the `io-code-block `directive specifying the overall language?

snooty/rstspec.toml
```diff
@@ -463,6 +463,45 @@ example = """.. code:: ${1:language}
...
 """
 
+[directive.io-code-block]
+help = """A block of code and its corresponding output to show the user."""
+content_type = "block"
+options.caption = "string"
+options.copyable = "boolean"
+options.class = "string"
+options.source = "uri"
+example = """.. io-code-block::
```
schmalliso, line 473: We've had a request to comment out the example for these directives so it won't shop up in the snippet view when folks are working in VSCode, since we'll be testing this in production and the content folks don't want to accidentally use it before they've had a chance to UAT.

TL:DR, please comment out the examples; we'll restore them in a couple of weeks once we've got the final LGTM.

schmalliso (approved):
LGTM, though Andrew should also review :)

snooty/parser.py
```diff
@@ -984,6 +1011,77 @@ def __make_child_visitor(self) -> "JSONVisitor":
         return visitor
 
 
+"""Validates that a given io-code-block directive has 1 input and 1 output
```
i80and, line 1014: Docstrings need to be inside of the function

i80and (commented):
One issue, then let's ship it!

i80and (approved):

PR #377, opened 2022-01-25
DOP-2696: Pass exceptions through WorkerLauncher rather than deadlock

i80and:


snooty/parser.py
```diff
@@ -1229,15 +1229,16 @@ def merge_diagnostics(
...
         """Run the postprocessor if and only if any pages have changed, and return postprocessing results."""
         postprocessor = self.postprocessor_factory()
         return self.worker.run(postprocessor)
 
     def flush_and_wait(self) -> PostprocessorResult:
-        result = self.flush().get()
-        if isinstance(result, util.CancelledException):
-            raise result
+        result, exception = self.flush().get()
+        if exception:
```
rayangler, line 1239: Just curious: is `exception` still going to be of type `CancelledException` or is it more generalized now? Was the specificity of `CancelledException` causing the program to not terminate?

rayangler (approved):
LGTM, can confirm that the postprocessor terminates now when hitting an assertion.

Just one question for my own understanding

snooty/parser.py
i80and: Correct: CancelledException is treated specially on the throwing and catching side, but the interface of the WorkerLauncher really doesn't need to treat it differently from any other exception.

PR #379, opened 2022-01-25
DOP-2649

i80and:


snooty/rstspec-consolidated.toml
```diff
@@ -0,0 +1,1781 @@
...
+help = """Link to the MongoDB contact page."""
+type = {link = "https://www.mongodb.com/contact%s"}
+
+[role.mdbacct]
+help = """Link to the MongoDB Account page."""
+type = {link = "https://account.mongodb.com/account%s"}
+
+[role.community-forum]
+help = "Link to the MongoDB Community Forums page."
+type = {link = "https://community.mongodb.com/%s"}
```
schmalliso, line 1136: this should apparently be `https://www.mongodb.com/community/forums/%s` now

snooty/specparser.py
```diff
@@ -1,12 +1,15 @@
 """Parser for a TOML spec file containing definitions of all supported reStructuredText
    directives and roles, and what types of data each should expect."""
 
+from __future__ import annotations
```
schmalliso, line 4: For my edification, what's this doing?

snooty/specparser.py
```diff
@@ -392,6 +397,16 @@ def resolve_value(key: str, inheritable: _T) -> _T:
         for key, inheritable in inheritable_index.items():
             resolve_value(key, inheritable)
 
+    @classmethod
+    def initialize(cls, text: str) -> None:
+        cls.SPEC = Spec.loads(text)
 
-GLOBAL_SPEC_PATH = util.PACKAGE_ROOT.joinpath("rstspec.toml")
-SPEC = Spec.loads(GLOBAL_SPEC_PATH.read_text(encoding="utf-8"))
+    @classmethod
```
schmalliso, line 404: Would it make sense to have an error returned if there spec.get results in no spec, or am I being overcautious?

schmalliso (approved):
One actual comment (though is sort of incidental to this PR, might as well fix it while we're here) and two questions :)

snooty/specparser.py
i80and: Python has a mechanism for adopting small language changes; before changes get turned on by default, you can import from `__future__` to opt into that feature.

In this case, it's upgrading past an annoying quirky interaction of type checking and the language where you can't reference a class within type annotations inside of that same class, because the class definition isn't yet complete. For example, this would be no good:

```
class A:
    def foo(self, other: A):
        pass
```

Instead, you would have to do:


```
class A:
    def foo(self, other: "A"):
        pass
```

This changes in Python 3.8, and the first form becomes valid. But because we're currently targeting Python 3.7, we have to do this import to opt into the Python 3.8 behavior.

snooty/specparser.py
i80and: Spec.get() could throw TomlDecodeError (if the TOML is invalid), LoadError (if the schema doesn't match what we expect), or some form of OSError if the built-in rstspec can't be read. The assertion should never trip (hence why it's an assertion: we're just proving a property to the type system).

We could conceivably catch these exceptions to print something nicer, but that's more important with external users.

snooty/rstspec-consolidated.toml
i80and: Done!

PR #380, opened 2022-01-26
(DOP-2600) Directive for cta-banner

gjchong25:


snooty/rstspec.toml
```diff
@@ -86,6 +86,15 @@ example = """.. warning:: ${1:Title}
    ${2:Warning block content}
 """
 
+[directive.cta-banner]
```
i80and, line 89: Could we put this into the `mongodb` domain? i.e. `[directive."mongodb:cta-banner"]`

snooty/rstspec.toml
```diff
@@ -4,7 +4,7 @@ version = 0
 [enum]
 alignment = ["left", "center", "right"]
 backlinks = ["entry", "top", "none"]
-banner_variant = ["info", "warning", "danger"]
+banner_variant = ["info", "warning", "danger", "cta-banner"]
```
i80and, line 7: To clarify, this permits:
```
.. banner::
   :variant: cta-banner
```

this is intentional? it's fine if it is, just want to check!

i80and (changes requested):
dynamite stuff

snooty/rstspec.toml
gjchong25: Nope, removed it!

snooty/rstspec.toml
gjchong25: Yep!

snooty/parser.py
```diff
@@ -608,6 +609,12 @@ def handle_directive(
 
             self.validate_and_add_asset(doc, argument_text, line)
 
+        elif name == "cta-banner":
+            if options is None:
+                self.diagnostics.append(
+                    MissingRequiredOption("cta-banner", "url", line)
```
rayangler, line 615: We may not need this diagnostic / options check. Omitting the url option results in a `DocUtilsParseError` and prints out:

```
ERROR(documents.txt:32ish): "cta-banner" requires the following options: url
```

rayangler (commented):
Just one thing from me but looks good!

snooty/parser.py
i80and: Oh, good catch! I forgot that we supported required options via the `required = true` option in rstspec

i80and (approved):

rayangler (approved):
LGTM!

PR #382, opened 2022-02-02
DOP-2719: Generalize cta-banner icon option

gjchong25:
This change modifies the `cta-banner` directive to allow it to take an LG icon as an option, so that other LG icons can be used if needed. The default icon if none is specified is the "Play" icon. 

[DOP-2719 JIRA Ticket](https://jira.mongodb.org/browse/DOP-2719)

### DIFF
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -646,8 +646,10 @@ example = """.. banner::
 [directive."mongodb:cta-banner"]
 content_type = "block"
 options.url = { type = "uri", required = true }
+options.icon = "string"
 example = """.. cta-banner::
    :url: https://university.mongodb.com/
+   :icon: ${1:Leafygreen icon}
 
    ${2:Banner Text content}
 """
--- a/snooty/test_parser.py
+++ b/snooty/test_parser.py
@@ -898,7 +898,7 @@ def test_cta_banner() -> None:
     project_config = ProjectConfig(ROOT_PATH, "", source="./")
     parser = rstparser.Parser(project_config, JSONVisitor)
 
-    # Test valid cta-banner
+    # Test valid cta-banner with specified icon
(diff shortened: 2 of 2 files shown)


snooty/test_parser.py
```diff
@@ -898,7 +898,34 @@ def test_cta_banner() -> None:
     project_config = ProjectConfig(ROOT_PATH, "", source="./")
     parser = rstparser.Parser(project_config, JSONVisitor)
 
-    # Test valid cta-banner
+    # Test valid cta-banner with specified icon
```
i80and, line 901: My general stance is we don't need tests for just rstspec changes!

snooty/rstspec.toml
```diff
@@ -646,8 +646,10 @@ example = """.. banner::
 [directive."mongodb:cta-banner"]
 content_type = "block"
 options.url = { type = "uri", required = true }
+options.icon = "string"
```
i80and, line 649: Should this be a closed set of values?

gjchong25: So the frontend handles the logic and error checking for dealing with whether or not the option being passed in is a valid one. If we wanted to have a closed set of values, we'd have to include all of the potential LG icons, and there are a lot.

snooty/test_parser.py
gjchong25: Oh gotcha!

snooty/rstspec.toml
rayangler: For reference, there are about 88 potential icons to choose from [here](https://www.mongodb.design/component/icon/example/) (for now). LG does allow us to import the list of possible values on the frontend, which makes it easier to handle there with our current version of the LG Icon package if we don't want to actively maintain the same set in our rstspec.

i80and (approved):

PR #385, opened 2022-02-15
DOP-2760: Add visibility option to output directive

gjchong25:
This change adds a visibility option to the output directive, which allows writers to specify whether or not they would like the output code snippet to be visible or hidden on default.


i80and (approved):
delightful!

PR #387, opened 2022-02-23
DOP-2742: Restore examples for io-code-block, input, and output directives in rstspec.toml

gjchong25:
[DOP-2742](https://jira.mongodb.org/browse/DOP-2742)

### DIFF
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -471,14 +471,14 @@ options.caption = "string"
 options.copyable = "boolean"
 options.class = "string"
 options.source = "uri"
-# example = """.. io-code-block::
-#    ${2::copyable: (bool)
-#    :caption: (string)
-#    :class: (string)}
-#    :source: ${1: URL to source code}
+example = """.. io-code-block::
+   ${2::copyable: (bool)
+   :caption: (string)
+   :class: (string)}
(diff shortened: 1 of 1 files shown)


i80and:
I'm not convinced that `input` and `output` should have full examples including the parent `io-code-block` -- I could see snippet behavior being a little counterintuitive if people trigger the `input` snippet while they're already inside of an `io-code-block`.

Spitballing: maybe have the `io-code-block` snippet *include* the input and output directive children already, and then have `input` and `output` snippets only contain themselves.

Does this concern make sense?

snooty/rstspec.toml
```diff
@@ -471,14 +471,23 @@ options.caption = "string"
...
-#    :caption: (string)
-#    :class: (string)}
-#    :source: ${1: URL to source code}
-
-#    ${0:code input/code output content}
-# """
+example = """.. io-code-block::
+   ${2::copyable: (bool)
+   :caption: (string)
+   :class: (string)}
```
i80and, line 477: Trailing curly brace?

i80and (approved):
Ideally we would test this to make sure it works as intended -- I'm not myself super familiar with the vscode/textmate snippet system. But that needn't block anything.

PR #396, opened 2022-04-14
DOP-2657: Add data_fields

i80and:
This permits `rstspec.toml` to list fields that may be included in `snooty.toml` under the `data` namespace. This is to be used for Nathan's asset fetching script.

### DIFF
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -1,3 +1,5 @@
+data_fields = ["source_page_template", "assets"]
+
 [meta]
 version = 0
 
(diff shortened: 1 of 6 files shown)


snooty/rstspec.toml
```diff
@@ -1,3 +1,5 @@
+data_fields = ["source_page_template", "assets"]
```
schmalliso, line 1: What's `source_page_template` ?

schmalliso (commented):
This looks good though I am also slightly confused 😁

snooty/rstspec.toml
i80and: This may change, but I wanted to start getting infrastructure in for communicating source URLs to the frontend :D

schmalliso (approved):

PR #400, opened 2022-05-20
Fix upsert logic, and add manifest test

i80and:


snooty/main.py
```diff
@@ -151,7 +151,9 @@ def on_update(
         if page.query_fields:
             document.update({"query_fields": page.query_fields})
 
-        self.handle_document(build_identifiers, page_id.without_known_suffix, document)
+        self.handle_document(
+            build_identifiers, page_id, fully_qualified_pageid, document
```
rayangler, line 155: Ah, so the reason for the error was that we were trying to pass in just the `page_id`, such as `search`, but the `MongoBackend` was trying to replace the `<project>/<user>/<branch>/search` document? Is my understanding on the error correct?

rayangler (commented):
Using the parser with this branch no longer causes the duplicate key error, but I see that the documents in our `documents` collection do not have the `created_at` field present. The `metadata` document does seem to retain the field though. 

It seems like this change was made previously [here](https://github.com/mongodb/snooty-parser/pull/366/files#diff-36f482f5f6f24dfcaf3606a52d62c6d31865ae22d0cb91b550797e5da0e7394dL206)?

snooty/main.py
i80and: Yep, exactly. Ideally we would actually have separate types to avoid conflating fully qualified IDs with regular IDs, but anyway

i80and:
> Using the parser with this branch no longer causes the duplicate key error, but I see that the documents in our `documents` collection do not have the `created_at` field present. The `metadata` document does seem to retain the field though.
> 
> It seems like this change was made previously [here](https://github.com/mongodb/snooty-parser/pull/366/files#diff-36f482f5f6f24dfcaf3606a52d62c6d31865ae22d0cb91b550797e5da0e7394dL206)?

Aha! Yes this is a regression since the original PR *removed* support for MongoBackend, then I never re-added the `created_at` field.

Anyway, added it back! Thanks for catching this!

rayangler (approved):
Changes LGTM! Parsing no longer causes the duplicate key error, and the `created_at` field is there for the documents!

PR #408, opened 2022-06-14
DOP-3064: Validate relative URLs in card directives

drew-beckmen:
### Ticket

[DOP-3064](https://jira.mongodb.org/browse/DOP-3064)

### Notes

[DOP-3039](https://jira.mongodb.org/browse/DOP-3039) added support for relative URLs in the `Card` component in the frontend. This PR adds validations to the parser to ensure that any relative URL passed as an option to the card directive points to a valid `.txt` file within the docs site being parsed.

The relative URLs can have leading or trailing slashes (or no slashes at all if the referenced file is not nested within the `source` directory). This aligns with the [expectations in the frontend](https://github.com/mongodb/snooty/blob/f57d8464374d62808a822d1cbeab6bb65d538ec9/tests/unit/Card.test.js#L20).

**Current Behavior**:

Running the parser when an invalid relative URL is passed in the card directive does not generate an error.

**New Behavior**: 

Running the parser with an invalid relative URL in the card directive generates an error with an explanation for the user.

Content:
```
   .. card::
      :headline: Launch & Manage MongoDB
      :cta: Launch & Manage MongoDB Relative URL Test
      :url: /launch-manage!!!!
      :icon: /images/icons/atlas.svg
      :icon-alt: MongoDB Atlas icon
```

Build:
```
➜  docs-landing git:(dop-3039-content-test) ✗ python3 -m snooty build .
INFO:snooty.main:Snooty 0.13.5 starting
ERROR(index.txt:47ish): Error opening /Users/drew.beckmen/Development/mongodb/docs-landing/source/launch-manage!!!!.txt: No such file or directory for relative path /launch-manage!!!!
```

Replacing `url: /launch-manage!!!!` with `url: /launch-manage` results in a clean build.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -952,7 +952,9 @@ class JSONVisitor:
         )
 
         if not target_path.is_file():
-            err_message = f"{os.strerror(errno.ENOENT)} for relative path {url_argument}"
+            err_message = (
+                f"{os.strerror(errno.ENOENT)} for relative path {url_argument}"
+            )
             self.diagnostics.append(CannotOpenFile(target_path, err_message, line))
         elif re.search(r"([\/])\1", url_argument) is not None:
             self.diagnostics.append(MalformedRelativePath(url_argument, line))
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -945,14 +945,12 @@ class JSONVisitor:
 
     def validate_relative_url(self, url_argument: str, line: int) -> None:
         """Validate relative URL points to page within current docs site"""
-        fullPath = util.filepath_given_relative_url(url_argument, self.project_config.source_path)
-        if not Path(fullPath).is_file():
+        fullPath = Path(
+            util.path_given_relative_url(url_argument, self.project_config.source_path)
+        )
+        if not fullPath.is_file():
             errMessage = f"{os.strerror(errno.ENOENT)} for relative path {url_argument}"
-            self.diagnostics.append(
-                CannotOpenFile(
-                    fullPath, errMessage, line
-                )
-            )
+            self.diagnostics.append(CannotOpenFile(fullPath, errMessage, line))
 
     def validate_doc_role(self, node: docutils.nodes.Node) -> None:
         """Validate target for doc role"""
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -944,13 +944,16 @@ class JSONVisitor:
             )
 
     def validate_relative_url(self, url_argument: str, line: int) -> None:
-        """Validate relative URL points to page within current docs site"""
-        fullPath = Path(
-            util.path_given_relative_url(url_argument, self.project_config.source_path)
+        """Validate relative URL points to page within current docs site.
+        URLs can be of the form /foo, foo, /foo/"""
+        relative_txt_path = Path(url_argument).with_suffix(".txt")
+        _, absolute = util.reroot_path(
+            FileId(relative_txt_path), self.docpath, self.project_config.source_path
         )
-        if not fullPath.is_file():
+
+        if not absolute.is_file():
             errMessage = f"{os.strerror(errno.ENOENT)} for relative path {url_argument}"
-            self.diagnostics.append(CannotOpenFile(fullPath, errMessage, line))
+            self.diagnostics.append(CannotOpenFile(absolute, errMessage, line))
 
     def validate_doc_role(self, node: docutils.nodes.Node) -> None:
         """Validate target for doc role"""
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -946,14 +946,13 @@ class JSONVisitor:
     def validate_relative_url(self, url_argument: str, line: int) -> None:
         """Validate relative URL points to page within current docs site.
         URLs can be of the form /foo, foo, /foo/"""
-        relative_txt_path = Path(url_argument).with_suffix(".txt")
-        _, absolute = util.reroot_path(
-            FileId(relative_txt_path), self.docpath, self.project_config.source_path
+        target_path = util.add_doc_target_ext(
+            url_argument, self.docpath, self.project_config.source_path
         )
 
-        if not absolute.is_file():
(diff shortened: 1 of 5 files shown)


snooty/util.py
```diff
@@ -83,6 +83,14 @@ def is_relative_to(a: Path, b: Path) -> bool:
         return False
 
 
+def path_given_relative_url(relative_url: str, project_root: Path) -> str:
+    """Returns absolute path to txt file specified by relative URL.
+    URLs can be of the form /foo, foo, /foo/"""
+    relative_url = relative_url[1:] if relative_url.startswith("/") else relative_url
+    relative_url = relative_url[:-1] if relative_url.endswith("/") else relative_url
+    return f"{project_root}/{relative_url}.txt"
```
rayangler, line 91: We could probably make use of the `reroot_path` util function to handle some of this logic. I think `pathlib` can handle some of the logic with the forward slashes as well

snooty/parser.py
```diff
@@ -940,6 +943,18 @@ def validate_and_add_asset(
                 CannotOpenFile(Path(image_argument), err.strerror, line)
             )
 
+    def validate_relative_url(self, url_argument: str, line: int) -> None:
+        """Validate relative URL points to page within current docs site.
+        URLs can be of the form /foo, foo, /foo/"""
+        relative_txt_path = Path(url_argument).with_suffix(".txt")
```
drew-beckmen, line 949: @rayangler I have one outstanding concern here which is that the comment `URLs can be of the form /foo, foo, /foo/` is not entirely correct. Because `Path` automatically strips any trailing slashes, if a writer put `/foo//` as the card URL, the parser would not complain. However, this would cause issues on the frontend as our users could potentially be routed to `<url>//`, which would always 404.

adding `url_argument.endswith("//")` to the if statement doesn't feel elegant to me, especially since the error could report an open file error on a valid path (since `absolute` will handle double slashes)

Any thoughts on a clean way to handle the double slash?

snooty/parser.py
```diff
@@ -940,6 +943,18 @@ def validate_and_add_asset(
                 CannotOpenFile(Path(image_argument), err.strerror, line)
             )
 
+    def validate_relative_url(self, url_argument: str, line: int) -> None:
```
i80and, line 946: Just chewing on this -- we have a *horribly* named `util.add_doc_target_ext()` function that might do what's needed here? It's invoked by `Project.validate_doc_role()`, which is pretty similar to this use case I think.

cc @rayangler

rayangler: That's a good point. I forgot that function existed 😅 Yeah, they do similar things so leveraging that might be better in this case

snooty/parser.py
rayangler: I'm unsure of the best way for this particular case. Do you have any expert Python knowledge on this @i80and ? I assume you could write a regex to check that there's no more than one consecutive `/` in the string at a time, but I'm not sure

drew-beckmen: I added the regex check to enforce the invariant specified in the comment & align with the expectations of the frontend

snooty/test_parser.py
```diff
@@ -119,6 +119,47 @@ def test_chapter() -> None:
...
+
+    page, diagnostics = parse_rst(
+        parser,
+        path,
+        CARD_CONTENT % (SOURCE_DIR + INVALID_URL),
+    )
+    assert len(diagnostics) == 1
+    assert isinstance(diagnostics[0], CannotOpenFile)
+
+
```
rayangler, line 162: Should we add a test case for a malformed relative path?

rayangler (commented):
I think things look good overall. Just one question/suggestion

snooty/test_parser.py
drew-beckmen: Ah yep - totally meant to do this, and it slipped my mind. Great flag - thank you!

snooty/parser.py
```diff
@@ -940,6 +944,21 @@ def validate_and_add_asset(
...
+        """Validate relative URL points to page within current docs site.
+        URLs can be of the form /foo, foo, /foo/"""
```
rayangler, line 949: Random thought, no change needed: I think ideally, users will use `/foo` since I believe that's the typical convention used for referring to other file names for other directives and roles. However, it does make sense to support the other forms as well since they're valid for navigation on the frontend

rayangler (approved):
LGTM! 

Not sure if @i80and has any further comments 👀

rayangler:
One last thing I remembered: idk if this is required, but it might be nice to update the expected inputs for `card`'s `url` option in the [rstspec.toml](https://github.com/mongodb/snooty-parser/blob/78631219292a1e24502a3dee5ed50c83501a0528/snooty/rstspec.toml#L755) file to explicitly include `path`

snooty/parser.py
```diff
@@ -940,6 +944,21 @@ def validate_and_add_asset(
...
+        target_path = util.add_doc_target_ext(
+            url_argument, self.docpath, self.project_config.source_path
+        )
+
+        if not target_path.is_file():
+            err_message = (
+                f"{os.strerror(errno.ENOENT)} for relative path {url_argument}"
+            )
+            self.diagnostics.append(CannotOpenFile(target_path, err_message, line))
+        elif re.search(r"([\/])\1", url_argument) is not None:
```
i80and, line 959: It's best to precompile regexes with `re.compile()` in the global scope: the python `re` module *does* maintain a cache, but when that cache reaches its maximum size (512 entries) the cache is wiped.

This normally isn't an issue, and shouldn't be in snooty, but in a larger application it means that dynamically constructed regexes can end up wiping out performance elsewhere unnecessarily.

i80and (changes requested):
Just one change, and then this will be good!

snooty/parser.py
drew-beckmen: TIL! thanks for sharing your Python knowledge! 🧠

i80and (approved):

PR #413, opened 2022-07-08
DOP-2986: Include branch name in manifest file

gjchong25:
Includes branch name in the manifest file as part of DOP-2986.


i80and (approved):
yeah this works for me!

PR #414, opened 2022-08-02
DOP-1166: add validation for icon role

seungpark:
[DOP-1166](https://jira.mongodb.org/browse/DOP-1166)

* add a constant file comprised of all icon css classes available in snooty
* add constant set module for comparison
* add verification for icon role that prepends font awesome css classname prefixes

TODO:
* have snooty-parser fetch snooty repo and parse [icon stylesheet](https://github.com/mongodb/snooty/blob/master/src/styles/icons.css) to auto update available css classes


snooty/parser.py
```diff
@@ -1002,6 +1007,23 @@ def validate_tabs_children(self, node: n.Directive) -> None:
...
         return
 
+    def validate_icon_role(self, node: docutils.nodes.Node) -> None:
+        """
+        Validate target for icon role
+        Checks for included icon file in root path
+        """
+        if not ICON_SET:
+            return
+        icon_name = node.__getitem__("target")
```
i80and, line 1017: `__getitem__` is the underlying method for the indexing operator `[]`. That is, this can just be `node["target"]`

snooty/parser.py
```diff
@@ -1002,6 +1007,23 @@ def validate_tabs_children(self, node: n.Directive) -> None:
...
+        possible_name_set = {
+            icon_name,
+            "fa-{}".format(icon_name),
```
i80and, line 1020: You can use f-strings for this: `f"fa-{icon_name}"`

snooty/parser.py
```diff
@@ -1002,6 +1007,23 @@ def validate_tabs_children(self, node: n.Directive) -> None:
...
 
+    def validate_icon_role(self, node: docutils.nodes.Node) -> None:
+        """
+        Validate target for icon role
+        Checks for included icon file in root path
+        """
+        if not ICON_SET:
+            return
+        icon_name = node.__getitem__("target")
+        possible_name_set = {
```
i80and, line 1018: I don't know the answer, just checking: are all of these cases always valid?

snooty/parser.py
```diff
@@ -1177,6 +1199,7 @@ def parse_inline(self, rst: str, lineno: int) -> MutableSequence[n.InlineNode]:
 
 def get_giza_category(path: PurePath) -> str:
     """Infer the Giza category of a YAML file."""
+    # returns first name in hyphenated path name
```
i80and, line 1202: This seems unrelated

snooty/diagnostics.py
```diff
@@ -696,3 +696,19 @@ def __init__(
...
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        icon_role: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(
+            f"Icon roles must be defined in .icon_names, supplied icon role is not defined: {icon_role}",
```
i80and, line 711: This message is a little implementation-specific and has a lot of stuff that writers may not care about. I might softly suggest
```
f"The Icon {icon_role} does not exist"
```
despite the lesser amount of information communicated.

snooty/parser.py
seungpark: they're not always valid. ie. there is `fa-archive` and `fa4-archive` but not `archive` so i went about with the set intersection. looking back seems like two prefixes missed ('charts-icon-xxxx' and 'mms-icon-xxxxx')

snooty/parser.py
seungpark: testing remnants. removed!

snooty/parser.py
seungpark: awesome. TIL py getters/setters

snooty/parser.py
```diff
@@ -1002,6 +1007,25 @@ def validate_tabs_children(self, node: n.Directive) -> None:
...
 
+    def validate_icon_role(self, node: docutils.nodes.Node) -> None:
+        """
+        Validate target for icon role
+        Checks for included icon file in root path
+        """
+        if not ICON_SET:
+            return
+        icon_name = node["target"]
+        possible_name_set = {
```
casthewiz, line 1018: non-blocking nitpick: We can name this something more contextual like `icon_name_formats` - since we're not exporting the var for usage, I'd be comfortable not hinting the type in the var name
```suggestion
        icon_name_formats = {
```

casthewiz (approved):
This LGTM with just a nitpick (nonblocking for merge imo)

snooty/parser.py
seungpark: good point. redundant without exposure. updated!

PR #415, opened 2022-08-08
DOP-3107: Validate list-table structure

i80and:


snooty/parser.py
```diff
@@ -973,9 +986,10 @@ def validate_doc_role(self, node: docutils.nodes.Node) -> None:
                 )
             )
 
-    def validate_list_table(
-        self, node: docutils.nodes.Node, expected_num_columns: int
-    ) -> None:
+    @staticmethod
+    def validate_list_table_item(
+        node: docutils.nodes.Node, expected_num_columns: int
+    ) -> Sequence[Diagnostic]:
```
rayangler, line 992: Just curious what the reason for turning this into a static method that returns a list of diagnostic(s) was. We have other validation methods as well that followed the same pattern as the previous iteration's. Is there a preference/ideal?

rayangler (approved):
Just one general question, but actual change LGTM

snooty/parser.py
i80and: instance methods drag around the entirety of `self`'s state, so you really don't know what it's going to do as a caller. This method wasn't actually *using* `self` for anything meaningful, so it was a little bit of a code smell.

That said, this entire *class* is a giant code smell, so this is kind of rearranging the deck chairs on the Titanic 😅

PR #416, opened 2022-08-26
Spelling suggestions for linking targets

i80and:
Implement Damerau-Levenshtein Distance spelling suggestions, with an edit distance of 2. Heuristics are used to severely limit the amount of entries in the target database we actually have to check, so even with large (non-adversarial) target databases this has low overhead.

However, each call is quite slow, so a pathological database could chew up a lot of time. It could be worth terminating if a set amount of time is wasted, and it might behoove us to copy the list of keys to check and release the lock quickly.

Example error:
```
ERROR(includes/steps-change-replica-set-wiredtiger.yaml:0ish): Target not found: "label:4.2-mmapv1-conf-option"
    Did you mean: std:label:4.2-mmapv1-conf-options
```

It would be better if the `did you mean` string were less... implementation-y and gave proper guidance about what should actually be written.

### DIFF
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -288,19 +288,24 @@ class UnknownSubstitution(Diagnostic):
     severity = Diagnostic.Level.warning
 
 
-class TargetNotFound(Diagnostic):
+class TargetNotFound(Diagnostic, MakeCorrectionMixin):
     severity = Diagnostic.Level.error
 
     def __init__(
         self,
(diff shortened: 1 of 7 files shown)


skerschb:
Andrew this is so cool! :)

i80and:
eeeee thank you @skerschb!!! 🥰 Hope you're doing well!

schmalliso (approved):

PR #417, opened 2022-08-26
DOP-1166: include all icon roles. parse icon classname from role + target

seungpark:
[improvements to icon validation](https://jira.mongodb.org/browse/DOP-1166)

previous solution: only checking :icon: role, and also has possibility of letting non existent icon targets to slip by since we are constructing a set of possible classnames

update: explicitly come up with icon classname from role name and target name, and validate this icon classname in set of valid icon classnames

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -292,7 +292,7 @@ class JSONVisitor:
                 self.state.append(role)
                 return
 
-            elif role_name.startswith('icon'):
+            elif role_name.startswith("icon"):
                 self.validate_icon_role(node)
 
             role = n.Role((line,), [], node["domain"], role_name, target, flag)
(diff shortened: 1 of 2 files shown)


snooty/parser.py
```diff
@@ -292,7 +292,7 @@ def dispatch_visit(self, node: docutils.nodes.Node) -> None:
                 self.state.append(role)
                 return
 
-            elif role_name == "icon":
+            elif role_name.startswith("icon"):
```
casthewiz, line 295: A part of me feels like this could become too permissive, but not a large concern and improbable that we'd add an `icon` prefix'd directive without wanting to add it to validation - so I'm fine with this and don't think we need to futureproof too much.

casthewiz (approved):

PR #418, opened 2022-09-06
Fix some icon issues

i80and:
* Fix MMS icon validation, removing a regression in docs-charts: https://github.com/10gen/docs-charts/blob/master/source/dashboard-refresh.txt#L83
* A couple of icon names have spaces that will probably break validation?


seungpark (approved):
good catch! thank you for the cleanup

i80and:
Thank YOU for doing this validation!! 🥳

PR #426, opened 2022-10-24
DOP-3307: Generate metadata for OpenAPI content pages

rayangler:
### Ticket

DOP-3307

### Notes

* Adds new `OpenAPIHandler` to the postprocessor to generate metadata for OpenAPI content pages. 
* Each OpenAPI content page should have only one `openapi` directive. Otherwise, an error occurs. This was originally planned to be handled by DOP-3308, but figured it was good to just add that logic to the handler.
* Update to `rstspec.toml` made.


snooty/postprocess.py
```diff
@@ -919,6 +919,57 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+
+        # source_type should be assigned in the parsing layer
+        source_type = node.options.get("source_type")
+        if not source_type:
+            return
+
+        source = ""
+        argument = node.argument[0]
+        if source_type == "local" or source_type == "atlas":
+            assert isinstance(argument, n.Text)
```
i80and, line 964: is this guaranteed to be true? i.e. can a writer subvert this?

snooty/postprocess.py
```diff
@@ -919,6 +919,57 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+            source = argument.get_text()
+        else:
+            assert isinstance(argument, n.Reference)
```
i80and, line 967: is this guaranteed to be true? i.e. can a writer subvert this?

snooty/postprocess.py
rayangler: This should be guaranteed to be true (unless there's some edge case I'm missing). The parser handles setting the `source_type` ([example](https://github.com/mongodb/snooty-parser/pull/426/files#diff-5e66e8539da7ed36bf30c7a27807d04eb3c8c9a0d78649afcdc4a90cb1871efeR715)), which is dependent on the argument and its node structure anyways. This logic in the postprocessor echoes that to grab the correct source without needing to go through similar parser logic to re-determine the argument's structure.

I also don't think a writer can subvert this. Since `source_type` isn't an exposed option in `rstspec`, the parser throws an error when attempting to mix and match `source_type` and the argument.

snooty/postprocess.py
rayangler: The comment for the line above should also answer this same question! Let me know if it doesn't though

snooty/postprocess.py
i80and: Yep makes sense! Thank you!

Could you put that in a comment? After that, lgtm!

rayangler: Done! https://github.com/mongodb/snooty-parser/pull/426/files#diff-d321f31a0d6e1148a3d4e4ec2fded3b660d822fcd629909c38fcc30dd3bc493eR963-R967

i80and (approved):

PR #428, opened 2022-11-02
DOP-3353: Change OpenAPI metadata source to be spec string

rayangler:
### Ticket

DOP-3353

### Notes

* It's more convenient to have the spec file's actual content as a string within the metadata. Since the parser layer already handles grabbing the content and turning it into a string, all the postprocessor has to do is grab it. This also avoids having to re-read the spec files across more than one of our tools (limiting it to the parser).

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -959,18 +959,15 @@ class OpenAPIHandler(Handler):
             return
 
         source = ""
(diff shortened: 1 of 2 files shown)


i80and (approved):
fewer type assertions make me MIGHTILY pleased!

PR #429, opened 2022-11-03
DOP-3200

seungpark:
[DOP-3200](https://jira.mongodb.org/browse/DOP-3200)

**status**: ready for review

changes include:
- parser takes in an external toc node marked with `${label} <|${project-name}|>` within a toc tree directive ie 
```
... toctree::

    /getting-started
    Atlas CLI <|atlas-cli|>                        <-- net new format for external node
    Inner Page Toc </inner-page>
```
- parser emits a warning when a external toc node is not listed under snooty.toml .associated_products
- parser emits an error when an external toc node is repeated within the doc repo

### DIFF
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -421,7 +421,9 @@ class MissingAssociatedToc(Diagnostic):
         end: Union[None, int, Tuple[int, int]] = None,
     ) -> None:
         super().__init__(
-            f"The toctree entry {expected_project} does not exist in project config associated projects. Removing this toctree entry",
+            f"""Detected an associated toctree entry at {expected_project} 
+            which does not exist in an associated_products entry within the snooty.toml. 
+            Removing this toctree entry.""",
             start,
             end,
         )
@@ -437,7 +439,7 @@ class DuplicatedExternalToc(Diagnostic):
         end: Union[None, int, Tuple[int, int]] = None,
     ) -> None:
         super().__init__(
-            f"The external toctree entry {duplicated_toc} is duplicated",
+            f"Detected a duplicated associated toctree entry at {duplicated_toc}. Removing this toctree entry.",
             start,
             end,
         )
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -410,6 +410,7 @@ class InvalidTocTree(Diagnostic, MakeCorrectionMixin):
     def did_you_mean(self) -> List[str]:
         return [".. ia::"]
 
+
 class MissingAssociatedToc(Diagnostic):
     severity = Diagnostic.Level.warning
 
@@ -425,6 +426,7 @@ class MissingAssociatedToc(Diagnostic):
             end,
         )
 
+
 class DuplicatedExternalToc(Diagnostic):
     severity = Diagnostic.Level.error
 
@@ -439,6 +441,7 @@ class DuplicatedExternalToc(Diagnostic):
             start,
             end,
         )
+
 
 class InvalidIAEntry(Diagnostic):
     severity = Diagnostic.Level.error
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -425,6 +425,21 @@ class MissingAssociatedToc(Diagnostic):
             end,
         )
 
+class DuplicatedExternalToc(Diagnostic):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        duplicated_toc: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(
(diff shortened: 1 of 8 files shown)


snooty/rstparser.py
```diff
@@ -944,14 +944,19 @@ def make_toc_entry(
...
         if match:
             title, target = match["label"], match["target"]
+            # pipelines denote project reference
+            if target.startswith("|") and target.endswith("|"):
+                ref_project = target[1:-1]
+                target = None
         else:
             target = child
 
         if not title and util.PAT_URI.match(target):
```
casthewiz, line 957: I think we'll want to add an `or` clause here for the second part of this conditional, since the `PAT_URI` regex of `^(?P<schema>[a-z]+)://` should be matching on `://`, which we won't have in the reference.

Maybe just `and (util.PAT_URI.match(target) or ref_project):`?
the `target.startswith("|") and target.endswith("|")` is also suitable as a boolean, if we want to assign to a var and reuse.

casthewiz (commented):
Just some small stuff, but overall looks good!

snooty/diagnostics.py
```diff
@@ -411,6 +411,38 @@ def did_you_mean(self) -> List[str]:
...
+    severity = Diagnostic.Level.warning
+
+    def __init__(
+        self,
+        expected_project: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(
+            f"The toctree entry {expected_project} does not exist in project config associated projects. Removing this toctree entry",
```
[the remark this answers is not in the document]
seungpark: done!

snooty/diagnostics.py
```diff
@@ -411,6 +411,38 @@ def did_you_mean(self) -> List[str]:
...
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        duplicated_toc: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(
+            f"The external toctree entry {duplicated_toc} is duplicated",
```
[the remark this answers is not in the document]
seungpark: done!

snooty/rstparser.py
seungpark: oh good catch! added the ref_project case for erroneous node

casthewiz (approved):
This looks good, but I'd also want to get @i80and 's opinion before merge since they're more well versed in the parser's conventions :)

snooty/postprocess.py
```diff
@@ -1658,6 +1659,7 @@ def find_toctree_nodes(
         node: Dict[str, Any],
         toc_landing_pages: List[str],
         visited_file_ids: Set[FileId] = set(),
+        external_nodes: List[object] = [],
```
i80and, line 1662: Setting default parameters like this tends to be a bit of a code smell, since only one list is ever instantiated:
```
def foo(a=[]):
...     a.append(1)
...     return a
... 
>>> foo()
[1]
>>> foo()
[1, 1]
>>> foo()
[1, 1, 1]
```

If that's *intended* behavior, we should probably document it :D

snooty/postprocess.py
```diff
@@ -1669,6 +1671,18 @@ def find_toctree_nodes(
...
             for entry in ast.entries:
                 toctree_node: Dict[str, object] = {}
+                if entry.ref_project:
+                    toctree_node = {
+                        "title": [n.Text((0,), entry.title).serialize()]
+                        if entry.title
+                        else None,
+                        "options": {"project": entry.ref_project},
+                    }
+                    if toctree_node in external_nodes:
```
i80and, line 1681: If the only methods we call on `external_nodes` are `__contains__()` and `append()`, a set would be more appropriate than a list.

I'm also uncomfortable with the nature of this data structure. It obviously works since everything should be builtin types, but having full dictionaries is harder to reason about and makes typing impossible in our current python version target (3.7; this would be possible in 3.8 and later).

I think I would prefer using a tuple `Tuple[Optional[str], str]` which is what this all collapses to I believe: `(entry.title, entry.ref_project)`

snooty/postprocess.py
i80and: Also see below about typing stuff

i80and (changes requested):

snooty/postprocess.py
i80and: I think this just shouldn't be a default parameter since the postprocessor can be run many times in one process (e.g. in the vscode extension).

snooty/postprocess.py
seungpark: very insightful comment! thank you. addressed typing and used a set to initiate (vs a default list)

i80and (approved):
Lovely! 🚀

PR #431, opened 2022-11-17
DOP-3298: Upgrade release Python to 3.11

i80and:
This PR splits the CI and Release tasks into two different actions: `CI` runs on `(macOS-x86-64,Linux-x86-64) X py3.8,py3.11)`, while `Release` runs on `(macOS-x86-64,Linux-x86-64) X py3.11)`

On my laptop (i7-1065G7, 4 cores), this results in a 14% reduction in time building 10gen/docs-mongodb-internal#888fad61c726440d07de6f399cec6da925156c75: 41.263s down to 35.496s, lowest time of 3 runs

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,59 @@
+name: Release
+
+on:
+  push:
+    tags:
+    - 'v[0-9]*'
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        platform: [ubuntu-20.04, macos-latest]
+    runs-on: ${{ matrix.platform }}
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.11'
+        architecture: 'x64'
+    - name: Run tests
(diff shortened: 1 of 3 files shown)


schmalliso:
This seems fine, but I don't really understand how we'll run something that uses python 3.11 when the computer we're running it on doesn't have python 3.11 ? But that might mostly mean I'm not a good candidate for this PR.

i80and:
> This seems fine, but I don't really understand how we'll run something that uses python 3.11 when the computer we're running it on doesn't have python 3.11 ? But that might mostly mean I'm not a good candidate for this PR.

So we use PyInstaller to build the bundles, which does a lot of hackery that, long story short, bundles Python 3.11 and our Python code together.

schmalliso:
@i80and Does that mean we don't actually need to install python3 in the worker pool Dockerfile?

i80and:
> @i80and Does that mean we don't actually need to install python3 in the worker pool Dockerfile?

Alas, `mut`.

Now we could conceivably do the same thing for mut, but I don't think the payoff is worth it right now.

schmalliso (approved):

PR #432, opened 2022-11-24
DOP-3326: add role to link to php.net

schmalliso:


i80and (approved):
Is this for compatibility with an existing corpus, or is this greenfield work?

schmalliso:
Our PHP library engineering friends are doing the content migrations for the PHP Lib docs and require this for their links to work :)

PR #437, opened 2023-01-05
DOP-3438: Remove devhub data paths

i80and:


snooty/test_devhub.py
```diff
@@ -1,130 +0,0 @@
...
-from .test_project import Backend
-from .types import BuildIdentifierSet
-from .util_test import ast_to_testing_string, check_ast_testing_string
-
-
-@pytest.fixture
-def backend() -> Backend:
-    backend = Backend()
-    build_identifiers: BuildIdentifierSet = {}
-    with Project(Path("test_data/test_devhub"), backend, build_identifiers) as project:
```
rayangler, line 17: Curious if we should delete the actual test data as well for `test_devhub`? I don't see any other references to it

rayangler (commented):
Current deletions seem like they should be okay. Just a question about deleting related test data (see comment).

Also, will we be removing devhub-specific directives from the rstspec, or is that better done separately?

snooty/test_devhub.py
i80and: Oop! Yes, probably, forgot about test_data 😅

i80and:
@rayangler I've made this more comprehensive! Thanks for catching what I missed!

PR #446, opened 2023-02-12
Cleanup: use a type for associated products rather than Dict[str, object]

i80and:


seungpark (approved):
this looks much cleaner! thanks for typing associated products @i80and !

PR #449, opened 2023-02-14
DOP-3503: Add OpenAPI Versioning support to parser

mmeigs:
### Ticket

[DOP-3503](https://jira.mongodb.org/browse/DOP-3503)

### Notes

Adds support for OpenAPI versions. Version data should be validated and passed as build metadata.

New openapi directive options called `versions` and `api-version` are implemented to allow the parser to source version data from APIx endpoints.

In-depth information on these changes and parameters [here](https://docs.google.com/document/d/1F2tmPnFM1Kj_gFZDAyWJUAO0M5L1Bd6xQ4pHTcmaI9M/edit#heading=h.bt6gg2flm4js).

Three thoughts and questions are provided below!

### DIFF
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -803,4 +803,4 @@ class InvalidVersion(Diagnostic, MakeCorrectionMixin):
         self.major_versions = major_versions
 
     def did_you_mean(self) -> List[str]:
-        return self.major_versions
+        return list(self.major_versions)
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -770,6 +770,21 @@ class IconMustBeDefined(Diagnostic):
         )
 
 
+class InvalidOpenApiResponse(Diagnostic):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(
+            "OpenAPI Versioning data is malformed. Please contact DOP team.",
+            start,
+            end,
+        )
+
+
 class InvalidVersion(Diagnostic, MakeCorrectionMixin):
     severity = Diagnostic.Level.error
 
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -791,7 +791,7 @@ class InvalidVersion(Diagnostic, MakeCorrectionMixin):
     def __init__(
         self,
         api_version: str,
-        major_versions: List[str],
+        major_versions: Sequence[str],
         start: Union[int, Tuple[int, int]],
         end: Union[None, int, Tuple[int, int]] = None,
     ) -> None:
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -780,7 +780,11 @@ class InvalidVersion(Diagnostic, MakeCorrectionMixin):
         start: Union[int, Tuple[int, int]],
         end: Union[None, int, Tuple[int, int]] = None,
     ) -> None:
-        super().__init__(f"""Invalid OpenAPI Version Option ({api_version}) specified in options""", start, end)
+        super().__init__(
+            f"""Invalid OpenAPI Version Option ({api_version}) specified in options""",
+            start,
+            end,
+        )
         self.major_versions = major_versions
 
     def did_you_mean(self) -> List[str]:
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -768,3 +768,20 @@ class IconMustBeDefined(Diagnostic):
             start,
             end,
         )
+
+
+class InvalidVersion(Diagnostic, MakeCorrectionMixin):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        api_version: str,
+        major_versions: List[str],
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(f"""Invalid OpenAPI Version Option ({api_version}) specified in options""", start, end)
+        self.major_versions = major_versions
+
+    def did_you_mean(self) -> List[str]:
+        return self.major_versions
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -652,8 +652,7 @@ class JSONVisitor:
             uses_rst = options.get("uses-rst", False)
             # The frontend will grab the file contents from Realm, using the argument as its key
             uses_realm = options.get("uses-realm", False)
-            # Versioning will be dependent on present option(s)
-            versions = options.get("versions", None)
+            # Versioning will be dependent on present api_version option
             api_version = options.get("api-version", None)
 
             if argument_text is None:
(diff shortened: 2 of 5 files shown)


snooty/parser.py
```diff
@@ -678,6 +680,11 @@ def handle_directive(
                     self.diagnostics.append(InvalidURL(line))
                 return doc
 
+            if api_version:
+                doc.options["source_type"] = "atlas"
+                doc.options["api_version"] = api_version
+                return doc
+
             if uses_realm:
```
mmeigs, line 688: I have left :uses-realm: as a supported option for backwards compatibility. Going forward it might not be necessary at all, but just leaving the note here!

snooty/postprocess.py
```diff
@@ -976,7 +983,59 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
             assert isinstance(argument, n.Reference)
             source = argument.refuri
 
-        self.openapi_pages[current_slug] = self.SourceData(source_type, source)
+        api_version = node.options.get("api_version", None)
+        resource_versions: Optional[List[str]] = None
+
+        # Fetch OpenAPI versioning data if options are present
+        if api_version and source == "cloud":
+            # Fetch latest git_hash for S3 versioning data
```
mmeigs, line 991: I have opted for the more efficient fetching option from the Tech Spec. It seems about twice as fast. Not a huge gain, but every bit counts these days. This might mean we do not need to support the :versions: option at all.

snooty/postprocess.py
```diff
@@ -976,7 +983,59 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+            git_hash_response = requests.get("https://cloud.mongodb.com/version")
```
mmeigs, line 992: There are two urls needed for this process. I am totally unaware of where to place these constants... I searched for other examples in the parser and really struck out. I believe most urls are passed in, maybe. Halp?

branberry: Should we create a new directory to store constants? it could be like `snooty/constants/urls.py` maybe

snooty/postprocess.py
```diff
@@ -976,7 +983,59 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+                    major_versions = []
+                self.context.diagnostics[fileid_stack.current].append(
+                    InvalidVersion(api_version, major_versions, node.start[0])
+                )
+                return
+
+            if api_version in version_data.keys():
+                resource_versions = version_data.get(api_version)
+            else:
+                resource_versions = []
```
rayangler, line 1032: (Nit) We could potentially reduce this block to:

```
resource_versions = version_data.get(api_version, [])
```

I believe this handles the same key-checking logic + default value behavior. If this isn't the desired behavior, feel free to ignore!

snooty/rstspec.toml
```diff
@@ -704,6 +704,8 @@ argument_type = ["path", "string", "uri"]
 options.uses-rst = "flag"
 options.uses-realm = "flag"
 options.preview = "flag"
+options.versions = "string"
```
rayangler, line 707: Should we remove this option if we're planning to keep the versions URL within the postprocessor?

rayangler (commented):
The metadata looks good! Just a few comments for now

snooty/postprocess.py
mmeigs: Totally

snooty/rstspec.toml
mmeigs: I was planning on supporting it unless we are truly striking it from the design. Are we ok with removing :versions: as an option for the openapi directive?

snooty/parser.py
```diff
@@ -678,6 +680,11 @@ def handle_directive(
                     self.diagnostics.append(InvalidURL(line))
                 return doc
 
+            if api_version:
+                doc.options["source_type"] = "atlas"
+                doc.options["api_version"] = api_version
```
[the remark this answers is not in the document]
mmeigs: Ah, that makes sense I believe. I can wait to translate the "case" until the postprocessor. Is that what you mean?
We use kebab case for the writers' rst, we use snake case for the metadata, and we use camel case for the UI.

snooty/postprocess.py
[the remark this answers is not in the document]
mmeigs: Oh, I had no idea! Very cool

snooty/parser.py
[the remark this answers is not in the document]
rayangler: Yup, this option might be redundant to declare here when we can just grab the same value in the postprocessor like so: `node.options.get("api-version")`. We can definitely just wait to save it as `api_version` in the metadata

snooty/rstspec.toml
rayangler: I think it should be okay to remove for now if there's no current use case for it. I didn't see any code that suggests we're using the option's value (please correct me if I'm wrong!), so as-is, I don't think the option really does anything

snooty/postprocess.py
rayangler: A new directory or file could work. @i80and, do you have any strong opinions here?

rayangler (approved):
LGTM!

snooty/diagnostics.py
```diff
@@ -768,3 +768,39 @@ def __init__(
...
+        )
+
+
+class InvalidVersion(Diagnostic, MakeCorrectionMixin):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        api_version: str,
+        major_versions: List[str],
```
i80and, line 794: nit: Usually better to use `Sequence` in function parameters rather than `List` unless you need mutation. Sequence can be any kind of... well, sequence type, and is immutable.

snooty/postprocess.py
```diff
@@ -976,7 +983,63 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+            # Fetch latest git_hash for S3 versioning data
+            try:
+                git_hash_url = "https://cloud.mongodb.com/version"
+                git_hash_response = util.HTTPCache.singleton().get(git_hash_url)
+                git_hash = str(git_hash_response, "utf-8")
+
+                version_url = f"https://mongodb-mms-prod-build-server.s3.amazonaws.com/openapi/{git_hash}-api-versions.json"
+                version_response = util.HTTPCache.singleton().get(version_url)
+                decoded = str(version_response, "utf-8")
+                data = yaml.safe_load(decoded)
```
i80and, line 1000: This might be a good place to use flutter!

```
@dataclass
@checked
class OpenAPIData:
    versions: Dict[str, List[str]]


data = check_type(OpenAPIData, yaml.safe_load(decoded))
```

This will take care of all of the type validation logic you currently have

snooty/diagnostics.py
mmeigs: Ah, just realized that did_you_mean is only accepting List[str] as input. How do we feel about leaving this? Or about changing did_you_mean to accept Sequences as well?

i80and: Ah, forgot about `did_you_mean()`. That interface probably shouldn't be mutable, but it's fine; just return `list(self.major_versions)`

snooty/postprocess.py
i80and: I wouldn't be opposed to a `snooty/constants.py` file. Ideally anything like this should be configurable in `snooty.toml` for theoretical outside users, but this entire functionality sounds pretty mongodb-specific.

Either way, no need to do that in this PR.

i80and (approved):

PR #452, opened 2023-02-15
DOP-3502: make associated_product.versions an optional field

seungpark:
In order to remove [this snooty config from cloud docs](https://github.com/10gen/cloud-docs/blob/master/snooty.toml#L1076), this is a pre-req to make versions an optional value.

Currently it serves no purpose. For the initial launch, all versions of the associated product that are active in repo_branches will be included. In the future, associated product version management should also happen within hapley / repo_branches to have a single source of truth, and not leave this "associated versioning" option to snooty configs.

Tested with doc repos with/without version specified, respectively produces:

image

vs
image

but these versions have no effect on the merged ToC outcome from the persistence module. image below shows resulting document with the versions configured. the persistence module adheres to the existence in repos_branches document (not the snooty.toml configuration)

(bottom is snooty_dev.metadata with _id 63ed517b6b68628ac28006ae)

image

### DIFF
--- a/snooty/types.py
+++ b/snooty/types.py
@@ -127,7 +127,7 @@ class BundleConfig:
 @dataclass
 class AssociatedProduct:
     name: str
-    versions: Optional[List[str]] = field(default_factory=list)
+    versions: Optional[List[str]] = field(default=None)
 
     def serialize(self) -> SerializableType:
         return {"name": self.name, "versions": self.versions}
--- a/snooty/types.py
+++ b/snooty/types.py
@@ -127,7 +127,7 @@ class BundleConfig:
 @dataclass
 class AssociatedProduct:
     name: str
-    versions: List[str]
+    versions: Optional[List[str]] = field(default_factory=list)
 
     def serialize(self) -> SerializableType:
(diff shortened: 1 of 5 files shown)


snooty/types.py
```diff
@@ -127,7 +127,7 @@ class BundleConfig:
 @dataclass
 class AssociatedProduct:
     name: str
-    versions: List[str]
+    versions: Optional[List[str]] = field(default_factory=list)
```
i80and, line 130: Optional is `T | None` -- do we actually want to accept None here, or if it's omitted, do we want an empty list?

i80and (approved):

snooty/types.py
seungpark: we actually want to remove it completely but the cloud docs already has it, so it will be an intermediate step to make it optional :T (if we do remove it completely, parser will complain about an unknown config)

so in short, we want to accept None here

i80and: Got it. All else being equal I think I prefer not to have this be Optional at all -- I don't like None's without a reason to have them. But since it sounds like this will never be used by any parser code, I think it's good to merge as-is :D

PR #453, opened 2023-02-16
DOP-3407

seungpark:
**Goal:** remove logic from parser that transforms specs into RST

tested with openapi directives found in [cloud-docs](https://github.com/10gen/cloud-docs/blob/master/source/reference/api-resources-spec.txt), [docs-app-services](https://github.com/mongodb/docs-app-services/blob/305812ca10b9bb47fc282ada1c8dcbe49987aef6/source/admin/api/v3.txt#L9), and [landing](https://github.com/mongodb/docs-landing/blob/9913d0772b32f687c2d07bcf2cf005c4d283d028/source/openapi/preview.txt)

respectively, the AST's produced:

image
image
image


snooty/parser.py
```diff
@@ -69,7 +69,6 @@
 from .gizaparser.nodes import GizaCategory
 from .icon_names import ICON_SET
 from .n import FileId, SerializableType, TocTreeDirectiveEntry
-from .openapi import OpenAPI
```
rayangler, line 72: We could probably delete `openapi.py` as well. I don't think we reference any of its related classes or methods anywhere else

seungpark: good catch! was looking for any other imports. removed

snooty/parser.py
```diff
@@ -654,15 +653,14 @@ def handle_directive(
                     )
 
         elif name == "openapi":
-            # Parsing should be done by the OpenAPI renderer on the frontend by default
-            uses_rst = options.get("uses-rst", False)
-            # The frontend will grab the file contents from Realm, using the argument as its key
+            # OpenAPI directive is parsed here and prepped for OAS module within autobuilder
+            # the module is responsible for building OpenAPI specs and AST, post parse
```
rayangler, line 657: Great comments for context

rayangler (approved):
LGTM!

i80and (approved):
We can actually now remove `fett` from pyproject.toml, I believe!

seungpark:
> We can actually now remove `fett` from pyproject.toml, I believe!

done! also had to update the resulting poetry.lock file on install

PR #454, opened 2023-02-28
Do not swallow OSError when adding file watches

i80and:
I found that tests were suddenly failing for no apparent reason. After investigation, it turns out that vscode had exhausted my system's inotify quota, but we were eating the relevant exception.

The exception handler in question was really only targeted at the "cannot open file" case, which was already handled elsewhere, but was acting as a very delayed "remove the file we can't read from our list of files" action.

I hate our PendingTasks system anyway, so I've made PendingFigure immediately handle removing the StaticAsset from the Page if it can't open the referenced file, and gotten rid of the now-pointless base class PendingTask.

### DIFF
--- a/snooty/page.py
+++ b/snooty/page.py
@@ -3,25 +3,46 @@ from pathlib import Path, PurePath
 from typing import List, Optional, Set
 
 from . import n
-from .diagnostics import Diagnostic
+from .diagnostics import CannotOpenFile, Diagnostic
 from .n import FileId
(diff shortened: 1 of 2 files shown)


rayangler (approved):

i80and:
I've lost the exact context around this PR, and I'm not really happy with it anyway.

I think the superior approach is to just remove all of our watching infrastructure -- the current implementation is a major layering violation, and we'd be better served in general just using the LSP built-in monitoring.

PR #456, opened 2023-03-06
Oxidize the damerau_levenshtein_distance() function

i80and:


i80and:
This was a fun POC, but I'm not sure it's the right direction.

PR #457, opened 2023-03-14
DOP-3531: Be more precise about start-before/end-after matching

i80and:


schmalliso:
Tis a thing of beauty.

PR #459, opened 2023-03-16
DOP-3530: update description for step directive in snooty.tmol

seungpark:
seems usage of steps was unclear. this attempts to clear up the difference between headings and paragraphs within steps

### DIFF
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -751,10 +751,12 @@ example = """.. short-description::
 
 [directive."mongodb:step"]
 help = "Make a single, numbered step."
-example = """.. step:: ${0: Step's headline string}
-${1: Step content}"""
-argument_type = "string"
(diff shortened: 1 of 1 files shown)


schmalliso:
I thought the argument was required and the content was optional -- is that not how we're actually handling things in reality?

seungpark:
> I thought the argument was required and the content was optional -- is that not how we're actually handling things in reality?

this is actually the way we are handling things, but it gets blurry when the argument is multi lines.

[here is an example of staged changes](https://docs-mongodbcom-integration.corp.mongodb.com/945ffeb659ba18f1626ca60f50e90d38860ce428/DOP-3530/test-landing/seung.park/master/)

the sample RST is [here](https://github.com/seungpark/test-repo/blob/DOP-3530/source/index.txt#L16-L47)

i80and (approved):
I don't see a problem with this!

PR #461, opened 2023-03-23
Cb/DOP-3375

caesarbell:
Adds a new marker for `card-group` for handling driver tiles.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,3 @@ htmlcov/
 dist/
 .docs
 .vscode
-
-# virtual environment
-venv
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ htmlcov/
 dist/
 .docs
 .vscode
+
+# virtual environment
+venv
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -59,8 +59,8 @@ ${4:   :align:
 """
 
 # Intended for use on the ecosystem landing page only
-# deprecated
 [directive."mongodb:drivers-index-tiles"]
+deprecated = true
 
 # Intended for use on the docs.mongodb.com/legacy landing site only
 [directive."mongodb:deprecated-version-selector"]
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -59,6 +59,7 @@ ${4:   :align:
 """
 
 # Intended for use on the ecosystem landing page only
+# deprecated
 [directive."mongodb:drivers-index-tiles"]
 
 # Intended for use on the docs.mongodb.com/legacy landing site only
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -10,4 +10,7 @@ htmlcov/
 .venv/
 dist/
 .docs
-.vscode
\ No newline at end of file
+.vscode
+
+# virtual environment
+venv
```
i80and, line 16: I'd rather this just go into your local gitignore since it's not part of any wider team workflow 😸  https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file

snooty/rstspec.toml
```diff
@@ -58,6 +59,7 @@ ${4:   :align:
 """
 
 # Intended for use on the ecosystem landing page only
+# deprecated
```
i80and, line 62: Do we want to raise a warning if this is used? If so, we can add a

`deprecated = true`

field to this directive

i80and (changes requested):
Formatting error -- make sure to run `make format test` before pushing 😸

.gitignore
caesarbell: @i80and awesome thank you for that. Going to make these changes.

snooty/rstspec.toml
caesarbell: @i80and as this flag raises a warning will it still parse correctly? This will still need to be useable as we need backward compatibility until `docs-ecosystem` latest version is released.

i80and: It'll parse perfectly fine, it'll just log a warning

i80and (approved):
If the added warning is okay, I'm good with this! I love changes that only touch config files

snooty/test_parser.py
```diff
@@ -141,6 +141,7 @@ def test_card() -> None:
 .. card-group::
    :columns: 3
    :layout: carousel
+   :type: drivers
```
seungpark, line 144: when i remove this option, i'm not seeing the `card_group_type` to be set to `default` here. is this intended? otherwise LGTM!

caesarbell:
> If the added warning is okay, I'm good with this! I love changes that only touch config files

The warning should be fine, as it is a good reminder to go back and remove this once all releases are out in the wild.

caesarbell:
> If the added warning is okay, I'm good with this! I love changes that only touch config files

The warning should be fine, as it is a good reminder to go back and remove this once all releases are out in the wild.

caesarbell:
> If the added warning is okay, I'm good with this! I love changes that only touch config files

The warning should be fine, as it is a good reminder to go back and remove this once all releases are out in the wild.

snooty/test_parser.py
caesarbell: @seungpark that was the intent, if the options aren't set, there it should be `undefined`, but the optional option could either be `default` or `drivers`

i80and (approved):

PR #462, opened 2023-03-23
DOP-2317: add support for hidden flag to readconcern and method rstobjects

schmalliso:


snooty/specparser.py
```diff
@@ -109,7 +62,7 @@ class RefRoleType:
 
 #: Spec definition of a role: this can be either a PrimitiveRoleType, or
 #: an object requiring additional configuration.
-RoleType = Union[PrimitiveRoleType, LinkRoleType, RefRoleType]
+RoleType = Union["PrimitiveRoleType", "LinkRoleType", "RefRoleType"]
```
i80and, line 65: I suspect without proof this might be causing the test failure

schmalliso: It's only failing on 3.8, which is why I didn't catch it when I ran `make test` locally, which is INTERESTING…

i80and: what happens if you remove the double quotes? In *theory* our `from __future__ import annotations` incantation at the top of the file should make that work

i80and (approved):
:shipit:

PR #468, opened 2023-04-20
DOP-3660: Add openapi-changelog directive

mmeigs:
### Ticket

DOP-3660

### Notes

Adds openapi-changelog directive and api-version option.

### DIFF
--- a/snooty/test_parser.py
+++ b/snooty/test_parser.py
@@ -3527,3 +3527,28 @@ def test_invalid_icon() -> None:
     )
     page.finish(diagnostics)
     assert len(diagnostics) == 0
+
+
+# def test_openapi_changelog() -> None:
+#     path = ROOT_PATH.joinpath(Path("test.rst"))
+#     project_config = ProjectConfig(ROOT_PATH, "", source="./")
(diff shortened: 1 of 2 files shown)


snooty/test_parser.py
```diff
@@ -3527,3 +3527,30 @@ def test_invalid_icon() -> None:
     )
     page.finish(diagnostics)
     assert len(diagnostics) == 0
+
+
+def test_openapi_changelog() -> None:
```
i80and, line 3532: We don't actually need tests for rstspec-only changes since there are no new code paths being introduced :D

i80and (approved):

snooty/test_parser.py
mmeigs: :) I was wondering! But I felt like better safe than sorry. I'll delete!

i80and:
@mmeigs do we want this to go out to writers immediately?

mmeigs:
> @mmeigs do we want this to go out to writers immediately or with our next release?

Oh, this is a brand new project! This is the very first ticket. This will be finished *hopefully* in about 6 weeks. So, not looking to have anyone use it. Not that they will even know about this directive until then...

PR #469, opened 2023-04-24
DOP-3661: disallow duplicate changelog directives

mmeigs:
Emits error on duplicate `openapi_changelog` directives on a page.

Also, does legwork to account for `openapi_changelog` directives in `parser.py` and `postprocess.py`

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -660,6 +660,7 @@ class JSONVisitor:
                     )
 
         elif name == "openapi":
+            print("openapi alone")
             # OpenAPI directive is parsed here and prepped for OAS module within autobuilder
             # the module is responsible for building OpenAPI specs and AST, post parse
             uses_realm = options.get("uses-realm", False)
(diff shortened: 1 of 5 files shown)


snooty/parser.py
```diff
@@ -712,6 +713,19 @@ def handle_directive(
                 )
                 return doc
 
+        elif name == "openapi-changelog":
+            # Version Changelog will be dependent on present api-version option
+            api_version = options.get("api-version", None)
+
+            if not argument_text == "cloud":
```
i80and, line 720: this should be `if argument_text != "cloud"`

snooty/parser.py
```diff
@@ -712,6 +713,19 @@ def handle_directive(
...
+                self.diagnostics.append(ExpectedPathArg(name, line))
+                return doc
+
+            if api_version:
+                return doc
+            else:
```
i80and, line 726: No need to put this in an `else`, since if the condition is true we immediately return.

i80and (changes requested):
looks good modulo two nits! we could make the diagnostic a little more helpful if you really want to dig into the parser a bit more, but I think the functionality is okay as is!

mmeigs:
> we could make the diagnostic a little more helpful if you really want to dig into the parser a bit more

Good idea! I've added a couple diagnostics and tests to beef it up! Let me know what ya think

i80and (approved):
Delightful! lgtm! :shipit:

PR #471, opened 2023-05-04
DOP-3663: validate api-version for changelog

mmeigs:
Very simple PR. This validates that the only acceptable value for the `api-version` for the `openapi-changelog` directive is "2.0"

I've only added this to account in the future for when there will be more versions. An additional check.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -56,7 +56,6 @@ from .diagnostics import (
     InvalidLiteralInclude,
     InvalidTableStructure,
     InvalidURL,
-    InvalidVersion,
     MalformedGlossary,
     MalformedRelativePath,
     MissingAssociatedToc,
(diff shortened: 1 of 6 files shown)


snooty/parser.py
```diff
@@ -725,6 +726,10 @@ def handle_directive(
                 return doc
 
             if api_version:
+                if api_version != "2.0":
```
i80and, line 729: we have a mechanism in rstspec.toml called `enum` to enumerate allowed values for an option; it's unfortunate that we don't currently allow using that up to directive arguments

i80and (approved):

branberry (approved):
LGTM!

snooty/parser.py
mmeigs: Oh coooool!!! I will switch! Thanks, I'll let you know when I've re-configured for a quick glance!

i80and: Oh no, you'd have to change some stuff in the parser -- it's a flaw that it doesn't work as one would expect 😅  This is okay as-is for now unless you actually want to go mucking around!

mmeigs: oh! Just saw this. I just pushed up using enums, and it seems to work?? But let me know if I should revert!

mmeigs: I also deleted an unused enum

i80and (approved):
oh! beautiful!!

PR #475, opened 2023-05-19
DOP-3688

seungpark:
**Goals:** 
- Introduce a new `.. facet::` directive
- Include a new post-process handler that traverses the page content attaches a `facets` property to the page, and strips removes the AST node.
- TDD - included test to test the above

### DIFF
--- a/snooty/n.py
+++ b/snooty/n.py
@@ -508,3 +508,10 @@ class Transition(Node):
 class Table(Parent[Node]):
     __slots__ = ()
     type = "table"
+
+@dataclass
+class Facet(Parent[Node]):
+    __slots__ = ("name", "values")
+    type = "facet"
+    name: str
+    values: List[str]
--- a/snooty/n.py
+++ b/snooty/n.py
@@ -509,6 +509,7 @@ class Table(Parent[Node]):
     __slots__ = ()
     type = "table"
 
+
 @dataclass
 class Facet(Parent[Node]):
     __slots__ = ("name", "values")
--- a/snooty/n.py
+++ b/snooty/n.py
@@ -508,11 +508,3 @@ class Transition(Node):
 class Table(Parent[Node]):
     __slots__ = ()
     type = "table"
-
-
-@dataclass
-class Facet(Parent[Node]):
-    __slots__ = ("name", "values")
-    type = "facet"
-    name: str
-    values: List[str]
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -53,6 +53,7 @@ from .diagnostics import (
     InvalidTocTree,
     InvalidVersion,
     MissingChild,
+    MissingFacet,
     MissingOption,
     MissingTab,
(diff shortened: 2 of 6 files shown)


snooty/postprocess.py
```diff
@@ -1547,6 +1547,44 @@ def _attach_doc_title(self, fileid_stack: FileIdStack, node: n.RefRole) -> None:
...
 
+class FacetsHandler(Handler):
+    """Builds page.facets depending on facets found on nodes on this page"""
+
+    # TODO: should be able to handle facet propagation from project/directory/page -> page in the future
+    # when taxonomy and function of facets (propagation) are decided
+    def __init__(self, context: Context) -> None:
+        super().__init__(context)
+
+        self.facets: Dict[str, Union[List[object], str]] = {}
```
i80and, line 1558: Hm. I'd like to see concrete types here rather than `List[object]` if we can do that, whether by using TypedDict on the easier side of things:
```
class FacetDef(TypedDict, total=False):
    name: str
```

or a dataclass:
```
@dataclass
class FacetDef:
    name: str
    children: List[FacetDef]
```

i80and (changes requested):

snooty/postprocess.py
seungpark: the taxonomy is TBD and i think this makes it hard for defining types beforehand, not knowing what attributes each level might hold. ie. `genres` might have subgenres(?) but `programming language` might have sub versions. ` would it be clearer to have the arbitrary fields under one `subfacets` field

snooty/n.py
```diff
@@ -508,3 +508,11 @@ class Transition(Node):
 class Table(Parent[Node]):
     __slots__ = ()
     type = "table"
+
+
+@dataclass
+class Facet(Parent[Node]):
```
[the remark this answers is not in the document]
seungpark: ah this was used previously, but realized we are removing facets in the AST before outputting them for snooty consumption. removed

snooty/postprocess.py
seungpark: discussed offline. made usage of `SerializedNode` to prevent `object` typing here. it is a little bit more loose than intended, but the [sequence of serializableTypes](https://github.com/mongodb/snooty-parser/blob/master/snooty/n.py#L49) fits the need here. can be more explicit if need be.

given the first approach above, we would have to change the output / test assertions from:

```
facets == {
            "genres": [{"name": "reference"}, {"name": "tutorial"}],
            "target_platforms": [{"name": "atlas", "versions": [{"name": "v1.2"}]}],
        }
```

to convert to a list:
```
facets == [
                     { 'name': 'genres', 'children': [{"name": "reference"}, {"name": "tutorial"}]},
                     { 'name': 'target_platforms', 'children': [{"name": "atlas", "versions": [{"name": "v1.2"}]}],
        }]
```

i80and (approved):
🚀

PR #476, opened 2023-05-19
DOP-3688 validation

seungpark:
**Goals:**
- be able to parse and hold in memory a taxonomy toml file to validate usage of new `.. facet::` directive
- validate uses of said facet, and display non-fatal warnings when values are incorrect

### DIFF
--- a/snooty/specparser.py
+++ b/snooty/specparser.py
@@ -25,7 +25,7 @@ from typing import (
 import tomli
 from typing_extensions import Protocol
 
-from . import tinydocutils, util
+from . import n, tinydocutils, util
 from .flutter import check_type, checked
 
 #: Types of formatting that can be applied to a role.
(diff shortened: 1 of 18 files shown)


snooty/specparser.py
```diff
@@ -262,6 +262,35 @@ def create_role(self) -> Role:
         )
 
 
+@checked
+@dataclass
+class TaxonomySpec:
+
+    TAXONOMY: ClassVar[Dict[str, Any]] = {}
```
i80and, line 269: Can we type this using `flutter.check_type()` and `flutter.checked`? `Any` is necessary sometimes, but shouldn't be used for greenfield development.

e.g.
```
@checked
@dataclass
class GenreDefinition:
    name: str

@checked
@dataclass
class TaxonamySpec:
    name: str
    genres: List[GenreDefinition]

# etc.

taxonomy = check_type(TaxonomySpec, tomli.loads(...)
```

snooty/specparser.py
```diff
@@ -262,6 +262,35 @@ def create_role(self) -> Role:
         )
 
 
+@checked
```
i80and, line 265: You only need `@checked` if you're going to be calling `flutter.check_type()`, which I don't see being called in this PR.

seungpark: added the check_type!

snooty/specparser.py
```diff
@@ -262,6 +262,46 @@ def create_role(self) -> Role:
...
+
+@checked
+@dataclass
+class TargetPlatformDefinition:
+    name: str
+    versions: List[FacetDefinition]
+
+@checked
+@dataclass
+class TaxonomySpec:
```
i80and, line 280: One last series of requests: Could we move this and its dependent classes to a new file titled `taxonomy.py`?

snooty/specparser.py
```diff
@@ -262,6 +262,46 @@ def create_role(self) -> Role:
...
+
+    genres: List[FacetDefinition]
+    target_platforms: List[TargetPlatformDefinition]
+
+    @classmethod
+    def get_taxonomy(cls) -> TaxonomySpec:
```
i80and, line 286: Right now this could get called quite a bit. I'd rather load the TaxonomySpec once and cache it. See what we do for Spec:

https://github.com/mongodb/snooty-parser/blob/master/snooty/specparser.py#L405-L413

i80and (changes requested):

snooty/specparser.py
seungpark: done! not as explicit (no need to call anything to load) but it will hold in memory 👍

snooty/specparser.py
seungpark: great call. moved to its own module

i80and (approved):
mahvelous!

PR #478, opened 2023-05-24
DOP-3729: Update Page.static_assets set in IncludeHandler

i80and:
This makes `static_assets` not a lie when a page includes another.


rayangler (approved):
Nice 😎

PR #480, opened 2023-05-30
Add kotlin drivers to rst spec

mongodben:
question to reviewers: is there anywhere else in the snooty repo(s) that we should add the kotlin driver?


schmalliso (approved):
This looks perfect for the tab definitions - we will need to add a logo to the front-end so it has a little picture next to it though. Can you file a DOP ticket and we'll take care of that?

PR #481, opened 2023-05-31
DOP-3746: Use Pyston on linux amd64

i80and:
```
CPython 3.11: 51.130s
Pyston:       41.841s
```

Pyston likely isn't a permanent solution, but provides a low-risk free performance bump in the medium-term.


rayangler (approved):
🚀

PR #482, opened 2023-05-31
DOP-3738: Remove asset hashing

i80and:
To avoid breaking any interfaces, we fill out the checksum fields with an escaped version of the asset's fileid. e.g.
```
[
    "documents/index.bson",
    "site.bson",
    "assets/images-compass-create-database.png",
]
```

There's no significant performance implication to this, as it turns out, but it DOES remove a lot of code and simplify the architecture.


snooty/main.py
```diff
@@ -135,7 +135,7 @@ def on_update(
             "ast": page.ast.serialize(),
             "source": page.source,
             "static_assets": [
-                {"checksum": asset.get_checksum(), "key": asset.key}
+                {"checksum": asset.safe_key, "key": asset.key}
```
rayangler, line 138: Since the new checksum will be some variation of the filename, will we need to consider the possibility of two repos/projects having the same image filename, but different image data, when pushing assets up to the db via the persistence module?

i80and: Oy, does the persistence module store images in a global namespace? That makes sense, but that probably means the persistence module would have to compute hashes itself

rayangler: Yup, all assets are currently pushed into the same collection with the checksum as the unique id.

Computing the hashes on the persistence module will also require us to ensure the pages that reference images in the `static_assets` field are referring to the correct images stored in the db. But it feels odd if the persistence module modifies the AST's pre-populated field to ensure the mapping is correct... This is particularly only a concern for the Snooty Data API to ensure it returns the correct images for the requested pages, but maybe this will be a non-issue based on how we decide to approach images with the API moving forward?

I thought about potentially leveraging the `project` and `branch` of the current build to help with uniquely storing and identifying images, but this would potentially result in many copies of the same images.

rayangler: Gotcha. If we really do want to proceed with removal of hashing logic from the parser, the persistence module can definitely take over. We can file a ticket accordingly 😄

Let me know if you ever want to brainstorm about this together or discuss anything

i80and:
We want to retain asset hashing.

PR #492, opened 2023-07-21
:sparkles: DOP-3865 adds canonical control from the .toml file

caesarbell:
This adds the `canonical` tag to the `snooty.toml` file for pages that are completely EOL'd. 

In those edge cases, we are not able to programmatically determine the canonical tag or point to their current version as there is none. To solve this we give the writer the ability to provide us the desired location. 

The logic is then handled in Snooty, as it grabs the canonical URL and applies it to the canonical tag. 

More info can be found [here](https://github.com/mongodb/snooty/pull/857).

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -1323,12 +1323,14 @@ class SubstitutionHandler(Handler):
                 current_paragraph.append(element)
             else:
                 if current_paragraph:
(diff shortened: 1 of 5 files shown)


seungpark (approved):
this LGTM. clear tests to verify changes 👍

rayangler (approved):

PR #494, opened 2023-08-01
[DOP-3836]: Cascading Facets Implementation

branberry:
### Ticket
[DOP-3836](https://jira.mongodb.org/browse/DOP-3836)

### Notes

At this time, this PR will only handle propagating `facets.toml` files, it will not handle validating those facets. Also, there are some type errors that are being ignored. This will be addressed in [another story](https://jira.mongodb.org/browse/DOP-3921) that will also deal with validating the `facets.toml` files.


snooty/types.py
```diff
@@ -241,6 +241,36 @@ def read(
...
 
+    @staticmethod
+    def load_facet_file(path: Path) -> Tuple[SerializedNode, List[Diagnostic]]:
+        diagnostics: List[Diagnostic] = []
+
+        try:
+            with path.open("rb") as f:
+                data = tomli.load(f)
+        except FileNotFoundError:
+            pass
```
seungpark, line 252: minor. should this be a pass for 404?

test_data/test_project_facets/snooty.toml
```diff
@@ -0,0 +1,13 @@
+name = "test_data"
+title = "MongoDB title"
+canonical = "https://mongodb.com/docs/mongocli/install"
+
+[substitutions]
+guides = "MongoDB Guides"
+
+[[associated_products]]
```
seungpark, line 8: minor. we can leave out associated products if we are testing just facets

test_data/test_project_facets/snooty.toml
```diff
@@ -0,0 +1,13 @@
+name = "test_data"
+title = "MongoDB title"
+canonical = "https://mongodb.com/docs/mongocli/install"
+
+[substitutions]
```
seungpark, line 5: remove substitutions for testing facets

seungpark: actually instead of adding this new projects, can we add facets.toml to existing test_project and make sure nothing fails? also can add new directive to `test_project/source/index.txt`

seungpark (approved):
largely LGTM.

i would either clean up unnecessary directives / substitutions from the new test data, or include the new facets tests in existing test_project. huge work on this @branberry !

PR #495, opened 2023-08-01
DOP-3875

seungpark:
[JIRA Issue](https://jira.mongodb.org/browse/DOP-3875)

Goal: Convert the [taxonomy doc](https://docs.google.com/document/d/1niV98yR0ZL9Hrdzj2NT7ili4VgAOvpO8kgacLi280-E/edit) to a toml configuration for validation. Each `genre`,  `targetProduct`, and `programmingLanguge` should be represented in taxonomy.toml. **PTAL** at comments in the toml file for names that may have been murky.

Notes:
- "versions" will be handled in a programmatic manner between mut indexing (generate manfest) and search transport server (save documents) to ensure consistency between search documents and `repos_branches` collection
- `programming_languages` have been verified against `search-staging.documents` ( see `unique_code_langs` view under `search-staging` collection)
- Design mockups for front end show different "category names" ie. `target_platforms` becomes `product`. This should be a quick change all references when it is finalized

### DIFF
--- a/snooty/taxonomy.toml
+++ b/snooty/taxonomy.toml
@@ -181,6 +181,7 @@ name = "kotlin"
 
 [[programming_languages]]
 name = "php"
+display_name = "PHP"
 
 [[programming_languages]]
 name = "python"
--- a/snooty/taxonomy.toml
+++ b/snooty/taxonomy.toml
@@ -56,6 +56,7 @@ display_name = "BI Connector"
 
 [[target_platforms]]
 name = "com"
+display_name = "Cloud Manager"
 
 [[target_platforms]]
 name = "c2c"
@@ -88,10 +89,6 @@ name = "mongocli"
 display_name = "MongoDB CLI"
 
 [[target_platforms]]
-name = "mongocli"
-display_name = "MongoDB CLI"
-
-[[target_platforms]]
 name = "visual-studio-extension"
 display_name = "MongoDB for VS Code"
 
@@ -112,7 +109,7 @@ name = "realm"
 
   # nestled under docs-realm 
   [[target_platforms.sub_platforms]]
-  name = "fluter"
+  name = "flutter"
   display_name = "Dart Flutter SDK"
 
   [[target_platforms.sub_platforms]]
--- a/snooty/taxonomy.toml
+++ b/snooty/taxonomy.toml
@@ -4,55 +4,221 @@ name = "reference"
 [[genres]]
 name = "tutorial"
 
+# trying to map to docs "name" property 
+# for target_platforms and sub_platforms
+# as much as possible
 [[target_platforms]]
 name = "atlas"
-[[target_platforms.versions]]
-name = "v1.2"
-[[target_platforms.versions]]
-name = "master"
+
+  [[target_platforms.sub_platforms]]
+  name = "atlas-app-services"
+  display_name = "App Services"
+
+  [[target_platforms.sub_platforms]]
+  name = "charts"
+  display_name = "Charts"
+
+  [[target_platforms.sub_platforms]]
+  name = "atlas-cli"
+  display_name = "Atlas CLI"
+
+  [[target_platforms.sub_platforms]]
+  name = "data-api"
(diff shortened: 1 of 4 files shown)


i80and (approved):

snooty/taxonomy.toml
```diff
@@ -4,55 +4,221 @@ name = "reference"
...
 [[target_platforms]]
 name = "atlas"
-[[target_platforms.versions]]
-name = "v1.2"
-[[target_platforms.versions]]
-name = "master"
+
+  [[target_platforms.sub_platforms]]
+  name = "atlas-app-services"
+  display_name = "App Services"
```
sarahemlin, line 15: I was thinking that for these subproducts, since they'll all also have the higher level "Atlas" tag, we could take atlas off the name.  Or is the name value coming from how we differentiate between them in the repos?

sarahemlin: Except for Atlas CLI, which doesn't make sense without Atlas in front.

seungpark: most names are an attempt to map to a snooty project name ie. [atlas-cli](https://github.com/mongodb/docs-atlas-cli/blob/master/snooty.toml#L1) this will be crucial in the future if we want to source other properties, ie. versions

sarahemlin: Makes sense, Seung!  I need to document the internal/external names so that both writers and stakeholders understand.

snooty/taxonomy.toml
```diff
@@ -4,55 +4,221 @@ name = "reference"
...
+name = "mongodb-analyzer" # 404
+display_name = "MongoDB Analyzer"
 
 [[target_platforms]]
 name = "mongocli"
-[[target_platforms.versions]]
-name = "v1.0"
+display_name = "MongoDB CLI"
+
+[[target_platforms]]
```
sarahemlin, line 90: Looks like mongocli is in here twice.

snooty/taxonomy.toml
```diff
@@ -4,55 +4,221 @@ name = "reference"
...
+
+  [[target_platforms.sub_platforms]]
+  name = "triggers"
+
+[[target_platforms]]
+name = "bi-connector"
+display_name = "BI Connector"
+
+[[target_platforms]]
+name = "com"
```
sarahemlin, line 58: I think this one is Cloud Manager, but maybe it's weird because it builds from the same repo as Ops Manager?  Do we need a display_name?

snooty/taxonomy.toml
```diff
@@ -4,55 +4,221 @@ name = "reference"
...
+[[target_platforms]]
+name = "realm"
+
+  [[target_platforms.sub_platforms]]
+  name = "cpp"
+  display_name = "C++ SDK"
+
+  # nestled under docs-realm 
+  [[target_platforms.sub_platforms]]
+  name = "fluter"
```
sarahemlin, line 115: Is the folder flutter rather than fluter?

snooty/taxonomy.toml
seungpark: good catch! added a Cloud Manager display_name

snooty/taxonomy.toml
seungpark: removed the dup!

snooty/taxonomy.toml
seungpark: thank you for the catch @sarahemlin !

snooty/taxonomy.toml
```diff
@@ -4,55 +4,218 @@ name = "reference"
...
 name = "java"
-[[target_platforms.versions]]
-name = "v4.3"
\ No newline at end of file
+
+[[programming_languages]]
+name = "kotlin"
+
+[[programming_languages]]
+name = "php"
```
sarahemlin, line 183: @seungpark should PHP have a display_name since the capitalization is non-standard?

sarahemlin (approved):
LGTM!

PR #496, opened 2023-08-03
Add new Realm SDK roles

krollins-mdb:
The `@realm/react` package and JS Web SDK both have new standalone reference docs. This PR adds roles for those references docs to make linking to them easier in the Realm docs.


krollins-mdb:
Closed because I got myself into some Git shenanigans and don't have the time to deal with it. I'll reopen based on the `latest` branch.

PR #497, opened 2023-08-03
Add new Realm sdk roles

krollins-mdb:
The @realm/react package and JS Web SDK both have new standalone reference docs. This PR adds roles for those references docs to make linking to them easier in the Realm docs.

Duplicate of #496, but based on `latest` branch.


krollins-mdb:
I have no idea why checks are failing. I re-ran them with the same result. Anything I need to do here?

schmalliso:
> I have no idea why checks are failing. I re-ran them with the same result. Anything I need to do here?

Don't worry about it - this branch is a bit of an odd one, but you should be good.

schmalliso (approved):

PR #499, opened 2023-08-07
DOP-3924: Cache assets including source constant substitutions

i80and:
With this change, with a no-op build of the server manual, we go from 285 misses to 0 misses, and the single-process `parse rst` time goes from 6.50s to 2.43s (as usual, minimum of three runs on the reference metric unit My Laptop).

image


branberry (approved):
LGTM!

PR #500, opened 2023-08-09
[DOP-3921]: Implement concrete Facet type

branberry:
### Ticket
[DOP-3921](https://jira.mongodb.org/browse/DOP-3921)

### Notes

This PR implements a new data class to represent the `Facet` data type. It also incorporates facet validation for the `facets.toml` file.

Updates were made in the postprocessor as well to handle the new `Facet` type.


snooty/types.py
```diff
@@ -101,6 +102,25 @@ class ParsedBannerConfig:
     node: n.Directive
 
 
+@dataclass
+class Facet:
+    category: str
+    value: str
+    sub_facets: Optional[List["Facet"]] = None
+
+    def __lt__(self, other: "Facet") -> bool:
```
branberry, line 111: These were added for testing purposes so that we can sort the `Facet` array

test_data/test_project/source/driver-examples/facets.toml
branberry, line 1: I moved these back into a separate project directory since they required me to update the `test_project.py` `test()` function every time a change was made here that are irrelevant to the `test()` function.

snooty/taxonomy.py
```diff
@@ -44,9 +44,10 @@ def get_taxonomy(cls) -> "TaxonomySpec":
     @classmethod
     def validate_key_value_pairs(cls, facet_str_pairs: List[Tuple[str, str]]) -> None:
         taxonomy_ref = asdict(cls.get_taxonomy())
+        facet_pairs = copy.deepcopy(facet_str_pairs)
```
i80and, line 47: Because tuples and strings are both immutable, you don't actually have to do a deepcopy! Just do `facet_pairs = facet_str_pairs[:]`

snooty/postprocess.py
```diff
@@ -1552,36 +1552,48 @@ def _attach_doc_title(self, fileid_stack: FileIdStack, node: n.RefRole) -> None:
...
-        if not self.target.get(node.options["name"]):
-            self.target[node.options["name"]] = []
-        target_list = self.target[node.options["name"]]
-        if isinstance(target_list, list):
-            target_list.append(facet_node)
+        facet_node = Facet(category=node.options["name"], value=node.options["values"])
+
+        if self.parent_stack:
+            parent = self.parent_stack[-1][0]
+            if isinstance(parent.sub_facets, list):
```
i80and, line 1570: sub_facets is an `Optional[list]` right? Much better if that's the case to just have
```
sub_facets = parent.sub_facets
if sub_facets is not None:
    sub_facets.append(facet_node)
```

snooty/parser.py
```diff
@@ -1597,12 +1597,19 @@ def propagate_facets(self) -> None:
         for base, _, files in os.walk(root):
             if "facets.toml" in files:
                 facet_path = Path(os.path.join(base, "facets.toml"))
-                curr_facet, diagnostics = self.config.load_facet_file(facet_path)
+                curr_facet, diagnostics = self.config.load_facets_from_file(facet_path)
+
+                if curr_facet is None:
+                    logger.warning(
```
i80and, line 1603: I think this actually sounds like a `logger.error()`

But would it be a diagnostic instead of just a logging message?

snooty/taxonomy.py
i80and: Also, perhaps it would be better for this method to take a `Sequence[]` than a `List[]` if we're not mutating it?

branberry: I do have to do a deep copy because `facet_str_pairs` is a `List` and we are calling `pop()`. This lead to a side effect where subsequent calls to this method would be missing additional values needed for validation.

snooty/parser.py
branberry: Yes, that's correct! It would be a diagnostic, and it should already be included in the result. I'll remove this log.

i80and (approved):
:shipit:

snooty/taxonomy.py
```diff
@@ -42,11 +41,14 @@ def get_taxonomy(cls) -> "TaxonomySpec":
         return cls.TAXONOMY_SPEC
 
     @classmethod
-    def validate_key_value_pairs(cls, facet_str_pairs: List[Tuple[str, str]]) -> None:
+    def validate_key_value_pairs(
+        cls, facet_str_pairs: Sequence[Tuple[str, str]]
+    ) -> None:
         taxonomy_ref = asdict(cls.get_taxonomy())
+        facet_pairs = list(facet_str_pairs[:])
```
i80and, line 48: Oh, if you need a list, then `list()` already does a copy, so this should just be `list(facet_str_pairs)`

PR #502, opened 2023-08-13
Update dependencies

i80and:
Should resolve https://github.com/mongodb/snooty-parser/security/dependabot/5

Mypy decided to change some type annotations in a way that aren't really possible to satisfy even though I believe our usage to be sound, together with some possible mypy correctness issues, requiring some code changes.


snooty/gizaparser/nodes.py
```diff
@@ -37,7 +39,10 @@
 from ..page import Page
 from ..types import EmbeddedRstParser, ProjectConfig
 
-_T = TypeVar("_T", str, object)
+if TYPE_CHECKING:
```
seungpark, line 42: was a bit confused if this was necessary but verified with [this doc](https://github.com/python/mypy/tree/e266cdf04f7ded2ee2da89b5912402df52a8e254/mypy/typeshed/stdlib/_typeshed) and testing locally with direct import throws error 

```
    from _typeshed import DataclassInstance
E   ModuleNotFoundError: No module named '_typeshed'
```

seungpark (approved):
LGTM 👍

snooty/gizaparser/nodes.py
i80and: Yeah, it's very very silly and I hope someday DataclassInstance gets added to the stdlib... somehow

PR #506, opened 2023-08-15
Split out Realm Kotlin SDK role to sync and base URLs

cbullinger:
This PR: 
- updates the existing `kotlin-sdk` role to point to the base library
- adds a new `kotlin-sync-sdk` role, which points to the sync library


schmalliso (approved):
This looks good to me. @cbullinger Can you double check if anyone is using the existing role before we merge/release since it's a potentially breaking change. Once we have the 👍🏻 I'll get this out.

cbullinger:
@schmalliso We should be good to go with this merge. I've updated everything on the `docs-realm` repo that would be affected by this change, and no one else has confirmed using this role in other repos. Thanks for your help!

PR #508, opened 2023-08-22
DOP-3946: Add cache test

i80and:


snooty/parse_cache.py
```diff
@@ -69,16 +82,30 @@ def get(
...
                 actual_blake2b = hashlib.blake2b(dep_path.read_bytes()).hexdigest()
             except OSError:
+                self.stats.misses += 1
                 raise CacheMiss()
 
             if dep_blake2b != actual_blake2b:
+                self.stats.misses += 1
                 raise CacheMiss()
 
+        self.stats.hits += 1
```
rayangler, line 92: Would we ever want to consolidate `CacheStats` with the existing way in which the parser tracks and logs `hits` and `misses`?

i80and: Why yes, probably! 😅 Good catch!

PR #511, opened 2023-08-29
DOP-3966: Make all internal path handling relative to the project source

i80and:
As demonstrated by the new code in `test_project.py::test_cache()`, caches are no longer path-dependent!

A notable change here is that since we now need a FileId for `snooty.toml`, that gets made with `FileId("../snooty.toml")` because FileIds are relative to the source directory.

### DIFF
--- a/snooty/test_project.py
+++ b/snooty/test_project.py
@@ -1,4 +1,6 @@
+import shutil
 import sys
+import tempfile
 import threading
 import time
 from collections import defaultdict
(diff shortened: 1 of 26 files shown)


snooty/test_project.py
```diff
@@ -346,3 +353,16 @@ def test_cache() -> None:
             project.build(1, False)
 
             assert project.cache.stats == CacheStats(hits=2, misses=0, errors=0)
+
+            # Ensure that the cache is valid even if it's copied to a different directory
```
schmalliso, line 357: SLICK

rayangler (approved):

PR #513, opened 2023-09-06
DOP-3878: Move propogate_facets to the postprocessor

i80and:
This ensures that our PageDatabase remains valid for caching purposes.


branberry (approved):
LGTM! Thanks for updating this

PR #514, opened 2023-09-06
DOP-4008: Ensure that on_diagnostics() and set_diagnostics() stay in sync

i80and:
The whole diagnostics system is kind of a mess, but this should improve things and help ensure the invariant that, at least for a regular build, the two diagnostics reporting methods stay synchronized.

### DIFF
--- a/snooty/util_test.py
+++ b/snooty/util_test.py
@@ -1,3 +1,4 @@
+import collections
 import contextlib
 import inspect
 import os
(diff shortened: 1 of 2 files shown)


snooty/util_test.py
```diff
@@ -199,7 +203,25 @@ def on_delete(self, page_id: FileId, build_identifiers: BuildIdentifierSet) -> N
         pass
 
     def flush(self) -> None:
-        pass
+        # Verify that on_diagnostics and set_diagnostics are in sync
```
rayangler, line 206: This is a good addition

rayangler (approved):

PR #517, opened 2023-09-07
DOP-3988

caesarbell:
Working on creating the test. Once that is figured out I will push it to be included in this PR. 

**Notes** 
The other landing directives in this PR are placeholders. Other PRs will update those directives.


i80and (approved):
Rad!

snooty/rstspec.toml
```diff
@@ -641,6 +641,53 @@ example = """.. card-group::
...
+"""
+
+[directive."landing:products"]
+content_type = "block"
+help = """Container to hold a group of products"""
+example = """.. landing:products::
+   ${0: Product content}
+"""
+
+[directive."landing:client-libraries"]
```
seungpark, line 663: does this need an argument? i see that we are passing a heading [here](https://github.com/schmalliso/docs-landing/blob/3099d7975ecfa8f1f018034b35ba7b8a00af5416/source/index.txt#L72)

this is getting parsed to:

```
"children": [
	{
		"type": "paragraph",
		"position": {
			"start": {
				"line": {
					"$numberInt": "71"
				}
			}
		},
		"children": [
			{
				"type": "text",
				"position": {
					"start": {
						"line": {
							"$numberInt": "71"
						}
					}
				},
				"value": "Client Libraries\n:cta: View all\n:url: "
			},
			{
				"type": "reference",
				"position": {
					"start": {
						"line": {
							"$numberInt": "71"
						}
					}
				},
				"children": [
					{
						"type": "text",
						"position": {
							"start": {
								"line": {
									"$numberInt": "71"
								}
							}
						},
						"value": "https://www.mongodb.com/docs/drivers/"
					}
				],
				"refuri": "https://www.mongodb.com/docs/drivers/"
			}
		]
	}
]
```

we would want this separately parsed as a separate heading from the content block, and the `View all` is being treated separately from the reference to drivers

snooty/rstspec.toml
```diff
@@ -641,6 +641,53 @@ example = """.. card-group::
...
+content_type = "block"
+help = """Title and call-to-action for drivers card"""
+example = """.. landing:client-libraries::
+   ${0: Library content}
+"""
+
+[directive."landing:introduction"]
+content_type = "block"
+help = """Make an Introcution"""
```
seungpark, line 672:
```suggestion
help = """Make an Introduction"""
```

snooty/rstspec.toml
```diff
@@ -641,6 +641,53 @@ example = """.. card-group::
...
+
+[directive."landing:client-libraries"]
+content_type = "block"
+help = """Title and call-to-action for drivers card"""
+example = """.. landing:client-libraries::
+   ${0: Library content}
+"""
+
+[directive."landing:introduction"]
+content_type = "block"
```
seungpark, line 671: similar here as above. [the heading](https://github.com/schmalliso/docs-landing/blob/3099d7975ecfa8f1f018034b35ba7b8a00af5416/source/index.txt#L24) is being treated as as a paragraph

```
{
  "type": "paragraph",
  "position": {
    "start": {
      "line": {
        "$numberInt": "23"
      }
    }
  },
  "children": [
    {
      "type": "text",
      "position": {
        "start": {
          "line": {
            "$numberInt": "23"
          }
        }
      },
      "value": "Developer Data Platform"
    }
  ]
}
```

seungpark (changes requested):
i think we need to specify some arguments here (for treating headings, or we can change the AST to specify headings inside the new directives instead of passing as arguments). see examples in comments below!

snooty/rstspec.toml
caesarbell: @seungpark yes that is correct, but those are just placeholders as this is separate from `DOP-3988`. I need to make those changes for `DOP-3989` as that is the ticket for the client library work. I needed to put something in the as a placeholder to allow the parser to work locally so I could build and test out the UI.

snooty/rstspec.toml
caesarbell: @seungpark same as above, this is a placeholder as well. This task is asked in a different ticket. I can update this all in this ticket, and update the ticket as well if you prefer.  I just used these as placeholders to get the parser to successfully run.

snooty/rstspec.toml
```diff
@@ -641,6 +641,53 @@ example = """.. card-group::
    ${0: Card content}
 """
 
+[directive."landing:product"]
+content_type = "block"
```
seungpark, line 645: similar here too. don't we want an argument to parse for heading? this is accepting a heading arg [in the doc repo](https://github.com/schmalliso/docs-landing/blob/3099d7975ecfa8f1f018034b35ba7b8a00af5416/source/index.txt#L51C25-L51C48)  but getting parsed as a paragraph and being [styled as a header](https://github.com/mongodb/snooty/pull/885/files#diff-9172bb560b5a007439e019059cc60578677f90ee536366b794f256e9bee055d5R30)

seungpark (approved):
this LGTM 👍 i would note the changes for other directives are temporary but tickets are in flight so NBD 👍

snooty/test_parser.py
```diff
@@ -132,6 +132,35 @@ def test_chapter() -> None:
...
+         This card was built with a relative URL.
+"""
+
+    page, diagnostics = parse_rst(
+        parser,
+        path,
+        PRODUCT_CONTENT,
+    )
+
+    page.finish(diagnostics)
```
seungpark, line 160: i would also validate the page.ast with `check_ast_testing_string`

caesarbell: @seungpark I may need to circle back on this one. I have a bit more I need to learn and understand, and I don't think this should block merging. I will definitely work on this though.

PR #520, opened 2023-09-11
DOP-3742: Refactor YAML handling entry point into its own class

i80and:


snooty/gizaparser/domain.py
```diff
@@ -0,0 +1,210 @@
...
+                continue
+
+            steps, text, parse_diagnostics = giza_category.parse(path, optional_text)
+            file_diagnostics.extend(parse_diagnostics)
+
+            def create_page(filename: str) -> Tuple[Page, EmbeddedRstParser]:
+                page = Page.create(
+                    giza_node.path,
+                    filename,
+                    text,
```
branberry, line 186: It looks like the `text` variable potentially holds the text information from the previous iteration. Is this intended?

branberry (commented):
Hi Andrew! Just have a few small questions so far. Currently still reviewing, so I'll most likely leave more.

snooty/gizaparser/domain.py
```diff
@@ -0,0 +1,210 @@
...
+        logger.debug("needs_rebuild: %s", ",".join(needs_rebuild))
+        for file_id in needs_rebuild:
+            file_diagnostics: List[Diagnostic] = []
+            try:
+                giza_node = giza_category.reify_file_id(file_id)
+            except KeyError:
+                logging.warn("No file found in registry: %s", file_id)
+                continue
+
+            steps, text, parse_diagnostics = giza_category.parse(path, optional_text)
```
branberry, line 179: Question: What is `optional_text` exactly? Is it the text data we want to include in the page? 

If my understanding is correct, every node that needs to be rebuilt will have the same `text` information in the page node that is created for it. Is this intended?

snooty/parse_cache.py
```diff
@@ -41,13 +42,34 @@ class CacheData:
...
 
+    def ingest_yaml(self, yaml_domain: gizaparser.domain.GizaYamlDomain) -> None:
+        self.yaml_nodes.clear()
+        for category_name, category in yaml_domain.yaml_mapping.items():
+            if category.reified_nodes is None:
+                continue
+
+            for node in category.reified_nodes.values():
+                source_hash = hashlib.blake2b(bytes(node.text, "utf-8")).hexdigest()
+                self.yaml_nodes[category_name][node.path] = (
```
branberry, line 65: I think I may be missing something here, but if we call `clear()` on this dictionary and attempt to assign a nested value, I would expect a `KeyError` to be raised. Apologies if this is not correct, my Python knowledge isn't the best 😅

branberry (commented):
Hi Andrew! Apologies for the delay. I have left a few more questions/comments.

snooty/parse_cache.py
i80and: It's a defaultdict! If a key is accessed which doesn't exist, the factory on line 46 (dict()) is called!

snooty/gizaparser/domain.py
i80and: We can either obtain file text from the filesystem or from the language server: `optional_text` allows us to manually specify the text of a given file, so that changes don't have to be saved to disk to be parsed!

snooty/parse_cache.py
branberry: Ah, that makes sense. Thanks!

snooty/gizaparser/domain.py
branberry: Gotcha! That makes sense to me. 

I feel like I may be mistaken, but it appears that if we provide `optional_text` here, then all of the nodes that need to be rebuilt will have the same `text` value provided.

branberry (approved):
Discussed with Andrew individually, and we can ignore my comments for now

PR #524, opened 2023-09-22
DOP-4043

seungpark:
**Context:** Currently, parser does not allow how to handle multiple facet values on a single ReST line.

This bug has been reported across different writers' PR's having warnings for multiple values on a single line:

See job logs for examples of `WARNING(data-api.txt:6ish): Facet specified is not a valid facet: programming_language:shell, json`
[cloud docs example here](https://github.com/10gen/cloud-docs/pull/4877)
[docs app services example here](https://github.com/mongodb/docs-app-services/pull/588)

**PR description:**
- parse `facet.values` for multiple string values separated by comma. construct multiple Facet objects from single line
- if a facet has a parent, assume this has only one value (as multiple children Facets cannot belong to same parent Facet)
- update unit tests to include multiple values in single line and add expected values

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -532,26 +532,25 @@ class JSONVisitor:
     def handle_facet(self, node: rstparser.directive, line: int) -> None:
         for value in node["options"]["values"].split(","):
             ref: Union[rstparser.directive, tinydocutils.nodes.Element] = node
-            for value in ref["options"]["values"].split(","):
-                single_value = value.strip()
-                try:
-                    facet_str_pairs: List[tuple[str, str]] = [
-                        (ref["options"]["name"], single_value)
-                    ]
+            single_value = value.strip()
+            try:
+                facet_str_pairs: List[tuple[str, str]] = [
+                    (ref["options"]["name"], single_value)
+                ]
 
-                    while ref.parent and ref.parent.get("name") == "facet":
-                        ref = ref.parent
-                        # parent facet with children can only have one value
-                        # no need to traverse in multiple directions
-                        facet_str_pairs.append(
-                            (ref["options"]["name"], ref["options"]["values"])
-                        )
-
-                    taxonomy.TaxonomySpec.validate_key_value_pairs(facet_str_pairs)
-                except KeyError:
-                    self.diagnostics.append(
-                        MissingFacet(f"{node['options']['name']}:{single_value}", line)
+                while ref.parent and ref.parent.get("name") == "facet":
+                    ref = ref.parent
+                    # parent facet with children can only have one value
+                    # no need to traverse in multiple directions
+                    facet_str_pairs.append(
+                        (ref["options"]["name"], ref["options"]["values"])
                     )
+
+                taxonomy.TaxonomySpec.validate_key_value_pairs(facet_str_pairs)
+            except KeyError:
+                self.diagnostics.append(
+                    MissingFacet(f"{node['options']['name']}:{single_value}", line)
+                )
 
     def handle_tabset(self, node: n.Directive) -> None:
         tabset = node.options["tabset"]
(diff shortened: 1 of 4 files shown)


snooty/parser.py
```diff
@@ -530,25 +530,27 @@ def dispatch_departure(self, node: tinydocutils.nodes.Node) -> None:
...
-                )
+        ref: Union[rstparser.directive, tinydocutils.nodes.Element] = node
+        for value in ref["options"]["values"].split(","):
+            single_value = value.strip()
+            try:
+                facet_str_pairs: List[tuple[str, str]] = [
+                    (ref["options"]["name"], single_value)
+                ]
+
+                while ref.parent and ref.parent.get("name") == "facet":
```
branberry, line 541: Hey Seung! I think there might be a bug here. If a child facet node has multiple values, on the second iteration, ref will no longer be defined and we won't enter the while loop. Because of this, the second value of the nested facet will only have a single tuple containing the nested facet category name and value, but not the parent. For example:

```rst
.. facet::
    :name: target_product
    :values: atlas

    .. facet::
        :name: sub_product
        :values: charts, atlas-cli
```

When checking the `sub_product` value `atlas-cli`, `ref` will be undefined because in the previous iteration we keep loop until we have `ref.parent`

branberry (changes requested):

snooty/parser.py
branberry: Just to confirm, I pulled your branch down and added a unit test for the facet that I specified in the last post. I did receive an error:

Users/brandonly/Documents/gitrepos/snooty-parser/snooty/test_parser.py::test_valid_facets failed: def test_valid_facets() -> None:
        path = FileId("test.rst")
        project_config = ProjectConfig(ROOT_PATH, "")
        parser = rstparser.Parser(project_config, JSONVisitor)
    
        page, diagnostics = parse_rst(
            parser,
            path,
            """
    
    .. facet::
        :name: target_product
        :values: atlas
    
        .. facet::
            :name: sub_product
            :values: charts, atlas-cli
    
    """,
        )
        page.finish(diagnostics)
>       assert len(diagnostics) == 0
E       AssertionError: assert 1 == 0
E        +  where 1 = len([MissingFacet('Facet specified is not a valid facet: sub_product:atlas-cli', (6, 0))])

snooty/test_parser.py:3678: AssertionError

branberry: One more comment, I think this might be a solution:

```python
        for value in node["options"]["values"].split(","):
            ref: Union[rstparser.directive, tinydocutils.nodes.Element] = node
```

We can reset the `ref` value for every iteration so that we don't mutate between iterations of the for loop. Tested it locally and it looks like the tests pass.

seungpark: thanks for the catch @branberry ! i added the small bit you suggested as the solution

i added the [test case to postprocess](https://github.com/mongodb/snooty-parser/pull/524/commits/4691464d71dd8640305c5d65b12bb4cdcb415b47#diff-bfac8a18bf662c13854408d2116c9293dc488fde36361670d1c7af27becdafa3L2896) (and had to make some changes) since this will be more verbose of actually constructing the output. PTAL !

snooty/postprocess.py
```diff
@@ -1651,23 +1651,43 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
         if not isinstance(node, n.Directive) or node.name != "facet":
             return
 
+        def get_children_total(
+            child: n.Node,
+        ) -> int:
+            if not isinstance(node, n.Directive) or node.name != "facet":
```
branberry, line 1657: NIT: Since we check if the node is a directive or not right before the function is declared, we can probably omit this check here.

branberry (commented):
Hey Seung! This looks good, just have one tiny nit

snooty/postprocess.py
seungpark: good catch! remnant of excessive checking

branberry (approved):
LGTM!

PR #528, opened 2023-10-04
DOP-4070: Automatically find a viable cache file from a given URL prefix

i80and:
We also gzip the cache files. Decompression of our largest file (Atlas) takes 100-200 milliseconds, but I judge it worthwhile for storage and bandwidth savings. In the future, switching to zstd would be worth it, but we don't need to think about that right now.

We also turn on pickle optimization by default to recoup some of that time cost.

### DIFF
--- a/snooty/parse_cache.py
+++ b/snooty/parse_cache.py
@@ -1,3 +1,4 @@
+import gzip
 import hashlib
 import logging
 import pickle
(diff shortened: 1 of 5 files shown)


rayangler (approved):

branberry (approved):
Looks good to me!

seungpark (approved):
LGTM 👍

PR #531, opened 2023-10-17
Upgrade CI to use CPython 3.12

i80and:


i80and:
Requires cx-freeze 6.16

PR #540, opened 2023-11-06
[DOP-3866]: Throw warning and ignore nested projects on detection within parser

branberry:
### Ticket 

[DOP-3866](https://jira.mongodb.org/browse/DOP-3866)

### Notes

This PR adds logic to handle nested projects. When parsing a project, we call the `get_file` method and provide the source directory as the root path. If we come across another directory containing a `snooty.toml`, we know that this is a nested project, and we do not traverse that directory.


snooty/test_util.py
```diff
@@ -75,6 +75,20 @@ def test_get_files() -> None:
     ]
 
 
+def test_get_files_nested() -> None:
+    # The test_data/nested_project paths test ho we handle nested projects within the monorepository.
```
mmeigs, line 79: how* (I think)

snooty/diagnostics.py
```diff
@@ -883,3 +883,15 @@ def __init__(
...
+class NestedProject(Diagnostic):
+    severity = Diagnostic.Level.warning
+
+    def __init__(
+        self,
+        nested_project: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ):
+        super().__init__(f"""Nested project detected: {nested_project}""", start, end)
```
mmeigs, line 897: Do we want more info here, as in letting the user know that the parser will not parse the nested project?

snooty/util.py
```diff
@@ -114,13 +119,22 @@ def get_files(
         # Preserve both the actual resolved path and the directory name
         dirs_set = dict(((base_resolved.joinpath(d).resolve(), d) for d in dirs))
 
+        nested_set = set()
+        for d_path, d_name in dirs_set.items():
+            dir_snooty_path = d_path / SNOOTY_TOML
```
seungpark, line 124: TIL about truediv

seungpark (approved):
this LGTM. good tests to showcase pulling files from nested project directories, as well as the diagnostic 👍

snooty/util.py
branberry: I actually didn't know about truediv until I googled it just now! Haha. I just saw that we were joining paths elsewhere in the codebase with a forward slash

PR #542, opened 2023-11-27
DOP-4141: Catch TOML parsing errors in snooty.toml

i80and:
Additionally make snooty.toml parsing errors generally fatal, rather than continuing with a blank project configuration.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -77,6 +77,7 @@ from .diagnostics import (
     UnexpectedIndentation,
     UnknownTabID,
     UnknownTabset,
+    UnmarshallingError,
 )
 from .gizaparser.nodes import GizaCategory
 from .icon_names import ICON_SET, LG_ICON_SET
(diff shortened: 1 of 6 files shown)


snooty/parser.py
```diff
@@ -102,6 +103,10 @@
 logger = logging.getLogger(__name__)
 
 
+class ProjectLoadError(Exception):
+    pass
```
seungpark, line 107: for my own understanding, exceptions will stop the process, as diagnostics will be appended to the log?

snooty/types.py
```diff
@@ -233,19 +233,21 @@ def open(cls, root: Path) -> Tuple["ProjectConfig", List[Diagnostic]]:
...
+                parsed_diagnostics.extend(cls.validate_data(result.data))
 
-                    return result, parsed_diagnostics
+                return result, parsed_diagnostics
             except FileNotFoundError:
                 pass
-            except LoadError as err:
+            except util.TOMLDecodeErrorWithSourceInfo as err:
+                diagnostics.append(UnmarshallingError(str(err), err.lineno))
+            except (LoadError, tomli.TOMLDecodeError) as err:
```
seungpark, line 250:
```suggestion
            except LoadError as err:
```

since we already caught `TOMLDecodeError` within `parse_toml_and_add_line_info`

seungpark (approved):
LGTM! tested with some dummy toml values. 🙌 easy to find the source error with the message

snooty/parser.py
i80and: Yep! This is raised *after* reporting diagnostics to the backend. So `main()` just catches this exception, flushes the backend so those diagnostics get reported, and then exits.

The second thing there is important: if the exception isn't caught, no diagnostics are reported. You just get an ugly backtrace.

seungpark: perfectimo! 👍

PR #544, opened 2023-12-14
[HOTFIX]: Add no caching flag for create-cache

branberry:


snooty/main.py
```diff
@@ -3,7 +3,7 @@
 Usage:
   snooty build [--no-caching] <source-path> [--output=<path>] [options]
   snooty watch [--no-caching] <source-path> [options]
-  snooty create-cache         <source-path> [options]
+  snooty create-cache [--no-caching]         <source-path> [options]
```
i80and, line 6: The space alignment here is just a little odd 😅 I think either all the `<source-path>`'s should be lined up, or this line should just have a single space

i80and (approved):

PR #546, opened 2024-01-02
0.15.2 Release

branberry:
Creating a new release commit to unblock [DOP-4170](https://jira.mongodb.org/browse/DOP-4170) by allowing the `--no-caching` flag to be provided to the `create-cache` command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [v0.15.2] - 2024-01-01
+## [v0.15.2] - 2024-01-02
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


rayangler (approved):
LGTM! 

Just wondering if there's a reason we're doing it through a PR rather than through the typical release process for the parser. Friendly reminder that a manual post-release bump might be needed!

branberry:
@rayangler Did this to make sure I was doing the right stuff when performing the release, haha! What is the typical release process? Do we commit these changes to master or is this done via creating the release itself?

rayangler:
@branberry The [HACKING.md](https://github.com/mongodb/snooty-parser/blob/main/HACKING.md#release-process) file has the instructions. I don't think it should be much different from what you currently have here, other than you'll need to do some things manually.

@i80and any thoughts?

PR #547, opened 2024-01-03
DOP-4207: pre determine image dimensions in parser

seungpark:
**Goal:** Add width and height properties to images and figures such that the AST contains these dimensions in order for snooty to lazy load images in the same dimension.

I added [this package](https://pypi.org/project/imagesize/) as a lightweight image size getter, but open to suggestion such as [PIL](https://pypi.org/project/pillow/).

Package has been installed via CLI command `poetry add imagesize@1.4.1` which updated pyproject.toml and poetry.lock subsequently.

[Example of staged changes](https://docs-mongodbcom-staging.corp.mongodb.com/master/docs/seung.park/DOP-4207/reference/write-concern/write-lifecycle/) with frontend branch

**Note:**
- added a `diagnostics` property to static_asset so that any errors while getting image dimensions will appended to the parser's diagnostics. ex: 
```
WARNING(core/csfle/reference/encryption-components.txt:0ish): WARNING! Intrinsic size for image not calculated at path /Users/seung.park/Desktop/doc-repos/docs-mongodb-internal/source/images/client-side-field-level-encryption-diagram.svg. Image may not be lazy loaded.
```

**Question:**
- how have we respected `width` and `scale` options of image directive? seems like from this [example of image directive usage](https://github.com/10gen/docs-mongodb-internal/blob/v7.0/source/reference/write-concern/write-lifecycle.txt#L18-L23), it is a existing bug to not respect this width ([HTML output](https://www.mongodb.com/docs/manual/reference/write-concern/write-lifecycle/))


snooty/diagnostics.py
```diff
@@ -916,3 +916,19 @@ def __init__(
...
+    severity = Diagnostic.Level.warning
+
+    def __init__(
+        self,
+        raw_path: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ):
+        super().__init__(
+            f"""WARNING! Intrinsic size for image not calculated at path {raw_path}. Image may not be lazy loaded.""",
```
i80and, line 931: Starting with `WARNING` is redundant: our loggers already include this information

snooty/types.py
```diff
@@ -16,6 +16,7 @@
     Union,
 )
 
+import imagesize  # type: ignore
```
i80and, line 19: It would be better to include a stub file in `stubs/`: e.g. https://github.com/mongodb/snooty-parser/blob/main/stubs/bson.pyi

snooty/types.py
```diff
@@ -64,8 +66,11 @@ class StaticAsset:
     fileid: FileId
     path: Path
     upload: bool
+    diagnostics: List[Diagnostic]
     _checksum: Optional[str]
     _data: Optional[bytes]
+    _width: Optional[float]
```
i80and, line 72: I think it would be better for these to be a single field, e.g. `_dimensions: Optional[Tuple[float, float]]`, since they're assigned together and we can't have a width without a height or vice versa

i80and (changes requested):
Nice! I've wanted this to happen for a while, so it's nice to see this done!

snooty/n.py
```diff
@@ -356,7 +356,7 @@ class Directive(Parent[Node]):
     domain: str
     name: str
     argument: MutableSequence["Text"]
-    options: Dict[str, str]
+    options: Dict[str, Any]
```
i80and, line 359: Can this be `object` instead of `Any` without causing things to blow up?

snooty/parser.py
```diff
@@ -159,6 +159,10 @@ def __call__(
         try:
             checksum = self.asset.get_checksum()
             options["checksum"] = checksum
+            if self.asset._width:
+                options["width"] = float(self.asset._width)
```
i80and, line 163: Attributes beginning with an underscore are by convention private. Either they shouldn't have a leading underscore, or `Asset` should have a method that fills out `options` itself. e.g. `options.update(self.asset.get_dimensions_dict())`.

snooty/n.py
seungpark: the mentioned blow up does happen. i can keep this to string and parse into a float on the front end

snooty/types.py
seungpark: i actually had it as a single field previously (single float field for [aspect_ratio](https://github.com/mongodb/snooty-parser/pull/547/commits/ea329a7fdad2ed2dcc945570e3d7efa0b5613a11) ), but converted to width and height. good point on a tuple 👍 will change

snooty/parser.py
seungpark: done! changed it to accessible attributes with Tuple(float, float)

snooty/types.py
seungpark: updated! but kept as two fields on the output directive to preserve output AST as width and height fields

snooty/types.py
seungpark: good catch! done

snooty/n.py
i80and: I actually don't have a strong opinion here. Moderate preference for sticking with a float until the frontend, I think? But your call!

seungpark: no strong preference! making it flexible with [str, Union[str, float]] was not pretty either

seungpark: discussed offline. making this `Any` or `Union[str, float]` still produces AST output of string. will keep as a string option

i80and (approved):

snooty/parser.py
```diff
@@ -159,6 +159,10 @@ def __call__(
         try:
             checksum = self.asset.get_checksum()
             options["checksum"] = checksum
+            dimensions = self.asset.dimensions
+            if dimensions is not None:
+                options["width"] = str(dimensions[0])
```
seungpark, line 164: @i80and seems like this will overwrite some input user data (ie. width option). validating this is None before overwriting in a follow up PR

PR #548, opened 2024-01-10
DOP-4207: preserve user width if specified

seungpark:
Previous PR overwrote user defined widths if they were supplied.

This PR addresses this so that images' and figures' option for width is preserved.

See updated test case to see the preservation in width/height ratio as well.

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -161,8 +161,14 @@ class PendingFigure(PendingTask):
             options["checksum"] = checksum
             dimensions = self.asset.dimensions
             if dimensions is not None:
-                options["width"] = str(dimensions[0])
-                options["height"] = str(dimensions[1])
+                user_width = options.get("width")
+                if user_width is None:
+                    options["width"] = str(dimensions[0])
+                    options["height"] = str(dimensions[1])
+                else:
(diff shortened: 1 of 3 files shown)


snooty/parser.py
```diff
@@ -161,8 +161,13 @@ def __call__(
...
             dimensions = self.asset.dimensions
             if dimensions is not None:
-                options["width"] = str(dimensions[0])
-                options["height"] = str(dimensions[1])
+                user_width = options.get("width")
+                if user_width is None:
+                    options["width"] = str(dimensions[0])
+                    options["height"] = str(dimensions[1])
+                else:
+                    width_num = float(re.sub(r"\D+", "", user_width))
```
i80and, line 169: I don't think it's a problem in this case, but *generally* speaking you want to be careful using the top-level ad-hoc `re` functions because they're implemented with a small capped cache of compiled patterns that just clears once ~100 entries are hit, so you can end up recompiling the same pattern many times

i80and (approved):

snooty/parser.py
seungpark: great point! moved to module level

PR #549, opened 2024-01-11
DOP-4207: remove warning for svgs

seungpark:
Goal: Remove warnings such as the following for images that do not need dimensions to preload in the browser.

```
WARNING(core/csfle/reference/encryption-components.txt:0ish): WARNING! Intrinsic size for image not calculated at path /Users/seung.park/Desktop/doc-repos/docs-mongodb-internal/source/images/client-side-field-level-encryption-diagram.svg. Image may not be lazy loaded.
```

Only the file formats that are defined in the constant are supported for image processing on the front end, and thus can be lazy loaded in the future.

### DIFF
--- a/snooty/types.py
+++ b/snooty/types.py
@@ -36,7 +36,7 @@ from .n import FileId, SerializableType
 FileSource = Union[Path, str]
 PAT_VARIABLE = re.compile(r"{\+([\w-]+)\+}")
 PAT_GIT_MARKER = re.compile(r"^<<<<<<< .*?^=======\n.*?^>>>>>>>", re.M | re.S)
-IMAGE_SIZING_EXT = [".png", ".avif", ".jpg", ".jpeg", ".webp"]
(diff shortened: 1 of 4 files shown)


snooty/types.py
```diff
@@ -36,6 +36,7 @@
 FileSource = Union[Path, str]
 PAT_VARIABLE = re.compile(r"{\+([\w-]+)\+}")
 PAT_GIT_MARKER = re.compile(r"^<<<<<<< .*?^=======\n.*?^>>>>>>>", re.M | re.S)
+IMAGE_SIZING_EXT = [".png", ".avif", ".jpg", ".jpeg", ".webp"]
```
i80and, line 39: Since this is only used for member existence, it should be a should be a set `{...}`

seungpark: ⚡ ~1/4 time for the win. updated!

i80and (approved):

PR #550, opened 2024-01-12
DOP-4225: flag images for lazy loading depending on 

seungpark:
**Goal:** Flag image directives for lazy loading, depending on their occurrence on the RST.

We can tweak the values for `min_section_depth` or `min_img_index`, but these were inspired from [this article for when to use lazy loading](https://web.dev/articles/lcp-lazy-loading?utm_source=lighthouse&utm_medium=devtools#testing_a_fix).

The front end portion is ready [here](https://github.com/mongodb/snooty/pull/981/files#diff-ffd2b22195ccbbae4128ae0f9aa63e3b223dec62b101275024fd4d21509f58e6R53). This will use Gatsby Image to lazy load components once they are in viewport

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -1743,7 +1743,7 @@ class FacetsHandler(Handler):
 
 
 class ImageHandler(Handler):
-    """Inspects the images on the page and appends a lazy loading option if the image is found below the fold.
+    """Inspects the images on the page and appends a lazy loading option if the image is considered to be below the fold.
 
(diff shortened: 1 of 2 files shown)


rayangler (approved):

i80and (approved):
If this is the heuristic we want, this looks good!

seungpark:
> If this is the heuristic we want, this looks good!

<3 open to tweaking values if we need to non-lazy load images earlier!

PR #553, opened 2024-01-23
DOP-3969: Raise a diagnostic for section structures that cannot fit into a tree

i80and:
Our approach here is to add a stack of "minimum heading levels" to our parsing memo. When we enter and exit a directive, we push and pop the current heading level, respectively.

This allows us to warn when we encounter a section with an unacceptably low section level.

### DIFF
--- a/snooty/test_parser.py
+++ b/snooty/test_parser.py
@@ -3738,6 +3738,7 @@ def test_inconsistent_levels() -> None:
     project_config = ProjectConfig(ROOT_PATH, "")
     parser = rstparser.Parser(project_config, JSONVisitor)
 
+    # Re-opening a section level is not permitted
     page, diagnostics = parse_rst(
         parser,
         path,
@@ -3760,6 +3761,7 @@ Query Result Format
     assert [type(d) for d in diagnostics] == [DocUtilsParseError]
     print(ast_to_testing_string(page.ast))
 
+    # Opening a new child level is fine
     page, diagnostics = parse_rst(
         parser,
         path,
@@ -3799,6 +3801,8 @@ Query Result Format
 """,
     )
 
+    # Ensuring for my own satisfaction that a diagnostic is raised if we try shenanigans in non-directive
+    # block contexts.
     page, diagnostics = parse_rst(
         parser,
         path,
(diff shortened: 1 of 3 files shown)


snooty/tinydocutils/states.py
```diff
@@ -1103,6 +1109,14 @@ def check_subsection(self, source: str, style: StyleKind, lineno: int) -> bool:
             if style.overline is not None:
                 memo.section_bubble_up_kludge = True
 
+            if level <= self.memo.minimum_level_stack[-1]:
+                self.parent.append(
+                    self.reporter.error(
+                        "Block content cannot contain sections that are sibling to the parent element's section level",
```
schmalliso, line 1115: So I understand what this message means, but I'm not sure our users will. Can we provide an example of when this happens and how to remediate?

snooty/test_parser.py
```diff
@@ -3730,3 +3730,133 @@ def test_duplicate_directive_option() -> None:
...
+    """DOP-3969"""
+    path = FileId("test.rst")
+    project_config = ProjectConfig(ROOT_PATH, "")
+    parser = rstparser.Parser(project_config, JSONVisitor)
+
+    page, diagnostics = parse_rst(
+        parser,
+        path,
+        """
+Query Result Format
```
schmalliso, line 3745: Can you add some comments to these describing why they're wonky or not wonky so it's clearer to scan?

schmalliso (changes requested):
Looks good to me structurally, just a few tiny suggestions to enhance clarity.

snooty/tinydocutils/states.py
i80and: It's tricky -- I would like to have a more detailed error message with an example, but historically we have short single-line error messages, with an exception for the rare suggested correction, and any explanation will be vertically very long.

What do you think about a link to a page to explain the issue in more detail, perhaps?

schmalliso: That sounds *great*!

i80and: https://github.com/mongodb/snooty-parser/wiki/Section-Hierarchy-Must-Be-Linear

How does this look?

snooty/test_parser.py
i80and: Done! ✍🏻

schmalliso (approved):

PR #556, opened 2024-01-24
DOP-4279: Parser identifies if an Instruqt lab is on the page

anabellabuckvar:
### Context

[DOP-4279](https://jira.mongodb.org/browse/DOP-4279)

Instruqt labs are currently embedded into the content of a docs page but instead, we want them to appear after a click of a button next to the page's title heading. To accomplish this, the frontend must know if the page has a lab within the page AST.

([Current display of Instruqt Labs](https://github.com/mongodb/docs/blob/eea48f019634747775072cbf6bdf8c9ff13b4b79/source/tutorial/getting-started.txt#L29))

Page ASTs have a page-level option that determines if a lab should be shown.
The instruqt directive has support for a new title option

### Notes
These changes were tested with tests added to test_processor.py. 
Only one Instruqt directive should be present per page; if there are two a diagnostic error will be produced.

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -1147,18 +1147,16 @@ class InstruqtHandler(Handler):
     def __init__(self, context: Context) -> None:
         super().__init__(context)
         self.has_instruqt_directive = False
-        self.instruqt_title = ""
 
     def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
         self.has_instruqt_directive = False
-        self.instruqt_title = ""
 
     def exit_page(self, fileid_stack: FileIdStack, page: Page) -> None:
         if not self.has_instruqt_directive:
             return
 
         else:
-            page.ast.options["instruqt_title"] = self.instruqt_title
+            page.ast.options["instruqt"] = self.has_instruqt_directive
 
     def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
         if not isinstance(node, n.Directive) or node.name != "instruqt":
@@ -1170,12 +1168,8 @@ class InstruqtHandler(Handler):
                     DuplicateDirective(node.name, node.start[0])
                 )
                 return
-
-            self.has_instruqt_directive = True
-            title = node.options.get("title", "")
-
-            if title:
-                self.instruqt_title = title
+            else:
+                self.has_instruqt_directive = True
 
 
 class IAHandler(Handler):
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -1157,7 +1157,7 @@ class InstruqtHandler(Handler):
         if not self.has_instruqt_directive:
             return
 
-        elif isinstance(page.ast, n.Root):
+        else:
             page.ast.options["title"] = [self.instruqt_title]
 
     def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
@@ -1168,7 +1168,7 @@ class InstruqtHandler(Handler):
             self.has_instruqt_directive = True
             title = node.options.get("title", "")
 
-            if len(title) != 0:
+            if not title:
                 self.instruqt_title = title
                 #  self.context.diagnostics[fileid_stack.current].append(
                 #     MissingOption() )
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -1141,6 +1141,39 @@ class OpenAPIChangelogHandler(Handler):
             return
 
 
+class InstruqtHandler(Handler):
+    """Identify if Instruqt directive is present on a page and add title as a page-level option if so"""
+
+    def __init__(self, context: Context) -> None:
+        super().__init__(context)
+        self.has_instruqt_directive = False
+        self.instruqt_title = ""
+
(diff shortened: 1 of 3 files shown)


snooty/postprocess.py
```diff
@@ -1141,6 +1141,39 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+
+    def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
+        if not isinstance(node, n.Directive) or node.name != "instruqt":
+            return
+
+        if isinstance(node, n.Directive) and node.name == "instruqt":
+            self.has_instruqt_directive = True
+            title = node.options.get("title", "")
+
+            if len(title) != 0:
```
i80and, line 1171: You can just say `if not title:` to be more idiomatic

snooty/postprocess.py
```diff
@@ -1141,6 +1141,39 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+
+    def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
+        self.has_instruqt_directive = False
+        self.instruqt_title = ""
+
+    def exit_page(self, fileid_stack: FileIdStack, page: Page) -> None:
+        if not self.has_instruqt_directive:
+            return
+
+        elif isinstance(page.ast, n.Root):
```
i80and, line 1160: `page.ast` should always be an `n.Root` instance

snooty/postprocess.py
```diff
@@ -1141,6 +1141,39 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
             return
 
 
+class InstruqtHandler(Handler):
+    """Identify if Instruqt directive is present on a page and add title as a page-level option if so"""
+
+    def __init__(self, context: Context) -> None:
+        super().__init__(context)
+        self.has_instruqt_directive = False
```
i80and, line 1149: What behavior do we want if a writer puts two instruqt directives on a page?

snooty/postprocess.py
i80and: Addendum: misread the code, it should be `if title:` :D

snooty/postprocess.py
```diff
@@ -1141,6 +1141,39 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+    def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
+        self.has_instruqt_directive = False
+        self.instruqt_title = ""
+
+    def exit_page(self, fileid_stack: FileIdStack, page: Page) -> None:
+        if not self.has_instruqt_directive:
+            return
+
+        else:
+            page.ast.options["title"] = [self.instruqt_title]
```
i80and, line 1161: I would rather the option name be e.g. `instruqt_title` rather than just `title`. And I'd prefer not to have the list wrapper around the value

snooty/postprocess.py
anabellabuckvar: Per @rayangler's advice, a diagnostic error will be produced

snooty/test_postprocess.py
```diff
@@ -2168,6 +2168,89 @@ def test_named_references() -> None:
         assert len(diagnostics) == 1
 
 
+def test_instruqt_directive() -> None:
```
i80and, line 2171: We should have a test with no title to prove that doesn't break!

i80and (approved):
:shipit: Nice!!

snooty/postprocess.py
```diff
@@ -1141,6 +1141,43 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+            return
+
+        else:
+            page.ast.options["instruqt_title"] = self.instruqt_title
+
+    def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
+        if not isinstance(node, n.Directive) or node.name != "instruqt":
+            return
+
+        if isinstance(node, n.Directive) and node.name == "instruqt":
```
rayangler, line 1167: (Nit) Looks like this conditional might be redundant given the conditional directly above

rayangler (approved):
LGTM! One optional nit

snooty/postprocess.py
```diff
@@ -1141,6 +1141,37 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+
+    def __init__(self, context: Context) -> None:
+        super().__init__(context)
+        self.has_instruqt_directive = False
+
+    def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
+        self.has_instruqt_directive = False
+
+    def exit_page(self, fileid_stack: FileIdStack, page: Page) -> None:
+        if not self.has_instruqt_directive:
```
i80and, line 1155: This structure is unidiomatic:
```
if <condition>
    return
else:
    <do things>
```

instead, just have
```
if <condition>:
    return

<do things>
```

i80and (changes requested):
Along with Raymund's request

i80and (approved):

PR #557, opened 2024-01-24
DOP-4181: Support multiple inline elements in a shared include substitution

i80and:


snooty/postprocess.py
```diff
@@ -126,10 +126,9 @@ def extract_inline(
     if (
         len(nodes) == 1
         and isinstance(node, n.Paragraph)
-        and len(node.children) == 1
-        and isinstance(node.children[0], n.InlineNode)
+        and all(isinstance(child, n.InlineNode) for child in node.children)
     ):
-        return [node.children[0]]
+        return cast(MutableSequence[n.InlineNode], util.fast_deep_copy(node.children))
```
schmalliso, line 131: For my edification, what's happening here?

schmalliso (approved):
Looks good, just curious about the return cast( etc. bit.

snooty/postprocess.py
i80and: We're cloning the children list instead of just taking the first child.

* The clone is unfortunate, but prevents unintentionally cascading mutation. *Ideally* we don't want nodes to exist in more than one place in the tree because then it's no longer a tree, it's a DAG. One reason Python is ill-suited to the parser is because cloning nodes is *vastly* more expensive than a simple memberwise copy should be. If you look at the fast_deep_copy() implementation, it's pickling and unpickling the data, which is insane but the most performant deep copy technique I could find.
* The cast is also unfortunate -- it should be sound, but our typing is kind of odd and Python doesn't give us to the tools to do this the right way.

PR #562, opened 2024-02-12
DOP-4370: In our `make package' target, test an actual build

i80and:


i80and:
Pudding proof: `CI/testPackage` on `macos-latest` fails: https://github.com/mongodb/snooty-parser/actions/runs/7877375736/job/21493408637?pr=562

PR #567, opened 2024-02-22
DOP-4368-B: remove `mongo-web-shell` from `latest`

mmeigs:
### Ticket

DOP-4368

### Notes

Removes `mongo-web-shell` from `latest` branch

### README updates

- - [ ] This PR does not introduce changes that should be reflected in the README.md and/or HACKING.md


snooty/rstspec.toml
```diff
@@ -779,13 +779,6 @@ help = "Make a kicker (a subheader that is placed on top of the main header)."
 example = """.. kicker:: ${1: string}"""
 argument_type = "string"
 
-[directive."mongodb:mongo-web-shell"]
-help = """An instance of the web version of a Mongo shell"""
-options.version = "mws_version"
```
rayangler, line 784: We should also be able to remove the `mws_version` enum, since I believe only this directive referenced it

rayangler (commented):
Pretty much looks good to me, but I think it'd be cool if we also remove one of the enums associated with this directive.

This is maybe a nit (I'm honestly not sure if we have a codified process), but I think we can make the base branch of the PR `main` and that would hopefully fix the CI failing. ~~It'll also be good to see the commit in the `main` branch directly. We can always sync up the `latest` branch afterwards.~~ (EDIT: Nevermind, I see some commits on `main` that were pulled in from `latest` and vice-versa, so I don't know if it really matters?)

snooty/rstspec.toml
mmeigs: Great callout

mmeigs:
@rayangler I believe that we don't expect the `latest` branch to pass tests. The only file that matters in this branch is the `rstspect.toml`, right? The rest of the branch is all old and messed up, I think. Which begs the question why I even changed any other files 🤷  . I found the last `latest` PR [here](https://github.com/mongodb/snooty-parser/pull/516) that doesn't pass either.

rayangler (approved):
Yes, you're right! For some reason I thought the CI test was required and blocked PR merge. (Also didn't realize you had a separate PR for `main`; I thought the B was for the parser and A was for the frontend 😆)

PR #574, opened 2024-03-15
DOP-4447: skip files that do not end with .txt 

seungpark:
### Ticket

DOP-4447

### Notes
- This change adds a check for file extension, in which case the parser will not save a bson to the zip file, therefore not handed to search manifests or snooty

### README updates

### DIFF
--- a/snooty/main.py
+++ b/snooty/main.py
@@ -42,14 +42,13 @@ from .n import FileId, SerializableType
 from .page import Page
 from .parser import Project, ProjectBackend, ProjectLoadError
 from .types import BuildIdentifierSet, ProjectConfig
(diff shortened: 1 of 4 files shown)


snooty/main.py
```diff
@@ -48,6 +49,7 @@
 logger = logging.getLogger(__name__)
 
 EXIT_STATUS_ERROR_DIAGNOSTICS = 2
+EXT_FOR_PAGE = ".txt"
```
i80and, line 52: We have this hard-coded in a lot of different spots

Could we actually move this to `util.py` so other files can use it?

i80and (approved):

snooty/main.py
seungpark: done! replaced a few places in postprocess which references this

i80and (approved):

PR #578, opened 2024-04-05
DOP-4479: Update github actions

i80and:
### Ticket

DOP-4479

### Notes

Test workflow succeeded in https://github.com/mongodb/snooty-parser/actions/runs/8573383377. Confirmed that a release was made with both assets attached.

### README updates

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,9 +29,9 @@ jobs:
       id: build_package
       run: make package
     - name: Upload package
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
-        name: packages
+        name: package-${{ matrix.platform }}
         path: dist/${{ steps.build_package.outputs.package_filename }}
 
   release:
@@ -40,9 +40,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Download
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
-          name: packages
+          path: packages
+          pattern: package-*
+          merge-multiple: true
       - name: Get environment
         id: environment
         run: |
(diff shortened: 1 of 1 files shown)


rayangler (approved):

.github/workflows/release.yml
```diff
@@ -40,17 +40,19 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Download
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
-          name: packages
+          path: packages
+          pattern: package-*
```
rayangler, line 46: Oh, actually, I have a question: Will the [path the Autobuilder uses](https://github.com/mongodb/docs-worker-pool/blob/24ec3e5c0a4b4f6c6a421862680d095063d4f816/Dockerfile#L51) to download the packages need to be updated?

i80and: Should just work! The zip file names aren't changing, and they look to be what I expect in the test pseudo-release I did*.

* which is slightly different from a real release, but only in a way that *shouldn't* matter.

.github/workflows/release.yml
```diff
@@ -29,9 +29,9 @@ jobs:
       id: build_package
       run: make package
     - name: Upload package
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
```
seungpark, line 32: (question) is there a reason we are building the parser in the test workflow (vs the release workflow)?

seungpark (approved):
this LGTM. thanks for providing the build log! wondering how you triggered a release job without pushing a tag?

.github/workflows/release.yml
i80and: The test workflow runs per OS in the matrix, while the release workflow only runs once and merges the resulting artifacts together. Our test processes do a build anyway to make sure the binaries continue to work

i80and:
> this LGTM. thanks for providing the build log! wondering how you triggered a release job without pushing a tag?

I added the branch to the condition for the workflow, and commented out the conditional for it needing to start with `v` 😹

PR #581, opened 2024-04-09
 DOP-4461: CI rstspec.toml 

i80and:
### Ticket

 DOP-4461

### Notes

Failed push tested in https://github.com/mongodb/snooty-parser/actions/runs/8625012846/job/23640921597?pr=581

### README updates


.github/workflows/test.yml
```diff
@@ -3,59 +3,27 @@ name: CI
...
-          token: ${{ secrets.GITHUB_TOKEN }}
-          draft: true
-          prerelease: true
+    - name: Fetch parser
+      run: |
+        mkdir -p downloaded_parser
+        curl -L "$(python3 ./tools/get_latest_parser_url.py)" -o downloaded_parser/snooty.zip
+        cd downloaded_parser && unzip snooty.zip
+    - name: Test parser
+      run: mkdir empty && ./downloaded_parser/snooty/snooty build --rstspec=snooty/rstspec.toml empty/
```
rayangler, line 29: (Just curious; might be out of scope) Would we ever want to run the parser tests with the rstspec file in this branch to flush out any inconsistencies between the rstspecs? Kind of an extreme example, but if `literalinclude` is removed from the latest rstspec, parsing this empty directory passes, but the rstspec is technically flawed.

i80and: Not a bad idea, actually!

i80and: I don't think that needs to block this, though. It can be a follow-up ticket if we decide to do that.

rayangler: Oh, yeah, for sure! Sorry, I should've approved before

rayangler (approved):

PR #582, opened 2024-04-10
update readme for release process

seungpark:
Removed the prereq for gpg as we sign with SSH keys instead !


i80and (approved):
Pedantically, you still need to set up your SSH key with git and github, but this is no longer a prerequisite.

PR #584, opened 2024-04-12
DOP-4543: Add new styles to generalize card content for programming language landing pages

rayangler:
### Ticket

DOP-4543

### Notes

* Please see related [frontend PR](https://github.com/mongodb/snooty/pull/1062) for more information, but the tl;dr is that this is done to help generalize centering content for cards so we don't have to rely on `:type: drivers` declared for non-drivers cards. 
* The `type` card option is being maintained for now to avoid immediate breaking changes, but we should probably remove it in favor of a more generalized style option.

### README updates


i80and (approved):
wunderbar!

seungpark (approved):

PR #590, opened 2024-05-04
DOP-4491: Ensure that cache data is invalidated when images change

i80and:
This removes the expensive_operation_cache, which is very old. The purpose is to avoid re-hashing image assets in the vscode use-case, but I don't think the benefit is worth having this weird interface around.

We should revisit in a post free-threading Python world when we can actually give workers access to shared state. That would also let us get rid of the entire pending task concept.

### Ticket

DOP-4491

### README updates


snooty/parser.py
```diff
@@ -144,32 +143,33 @@ def verify(self) -> None:
...
+    ) -> None:
         super().__init__(node)
         self.node: n.Directive = node
         self.asset = asset
+        self.dependencies = dependencies
 
     def __call__(
         self, diagnostics: List[Diagnostic], project: ProjectInterface
     ) -> None:
         """Compute this figure's checksum and store it in our node."""
```
seungpark, line 160: minor. can we update the comments here?

snooty/test_project.py
```diff
@@ -417,6 +391,65 @@ def test_cache() -> None:
                     )
 
 
+def test_image_invalidation() -> None:
```
seungpark, line 394: awesome! TIL programmatically testing the cache. previous cache is not used for changed image 👍

seungpark (approved):
thank you for the explanation, much appreciated! will have to fix this [minor svg file](https://github.com/10gen/docs-k8s-operator/blob/master/source/images/kubernetes-operator.bakedsvg.svg?short_path=dfbf22d) for ([this error](https://mongodb.slack.com/archives/C018UTGP23T/p1711377002896719) is actually correct in that it can't compute intrinsic size, but now previous cached image will be updated 🙏  )

snooty/parser.py
i80and: oop good eye!

PR #594, opened 2024-05-08
DOP-4441: Improve snooty.clickInclude command in VSCode extension

mayaraman19:
### Ticket

DOP-4441

### Notes

Adds a method to get original YAML file paths for rST files that have been converted from YAML files to support the [clickInclude](https://github.com/mongodb/snooty-vscode/blob/main/src/extension.ts#L102) function in the VSCode extension.

### README updates

### DIFF
--- a/snooty/language_server.py
+++ b/snooty/language_server.py
@@ -430,9 +430,9 @@ class LanguageServer(pyls_jsonrpc.dispatchers.MethodDispatcher):
             return str(resolved_target_path)
         elif resolveType == "directive":
             filePath = str(self.project.config.source_path) + fileName
-            uwu = self.project.fileid_to_source_path(Path(filePath))
-            finalFile = self.project.config.source_path / uwu
-            return str(finalFile)
+            fileId = self.project.get_fileid(Path(filePath))
+            finalFileId = self.project.config.source_path / fileId
+            return str(finalFileId)
 
         else:
             logger.error("resolveType is not supported")
--- a/snooty/language_server.py
+++ b/snooty/language_server.py
@@ -433,7 +433,7 @@ class LanguageServer(pyls_jsonrpc.dispatchers.MethodDispatcher):
             a YAML file. We want to get its original file path in that case."""
 
             # Strip the first slash from fileName so the / operator doesn't mess up :|
-            stripped_file_name = fileName[1:]
+            stripped_file_name = fileName.lstrip("/")
             if fileName.endswith("rst"):
                 file_path = self.project.config.source_path / stripped_file_name
                 file_id = self.project.get_fileid_from_ast(file_path)
--- a/snooty/language_server.py
+++ b/snooty/language_server.py
@@ -431,14 +431,16 @@ class LanguageServer(pyls_jsonrpc.dispatchers.MethodDispatcher):
         elif resolveType == "directive":
             """If the filename has a .rst extension, it might be converted from
             a YAML file. We want to get its original file path in that case."""
-            if ".rst" in fileName:
-                filePath = str(self.project.config.source_path) + fileName
-                fileId = self.project.get_fileid(Path(filePath))
-                finalFileId = self.project.config.source_path / fileId
-                return str(finalFileId)
-            else:
-                return str(self.project.config.source_path) + fileName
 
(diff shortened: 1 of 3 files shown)


snooty/parser.py
```diff
@@ -1527,7 +1527,7 @@ def __init__(
         self.expensive_operation_cache: Cache[FileId] = Cache()
         self.backend.on_config(self.config, branch)
 
-    def get_page_ast(self, path: Path) -> n.Node:
+    def get_page_ast(self, path: Path) -> n.Root:
```
mayaraman19, line 1530: these `n.Root` changes were made so the tests don't fail, since the `n.Node` class does not have the `fileid` option so tests would automatically fail. I'm not 100% sure if converting this to `n.Root` is the correct assumption.

i80and: Yeah, this is the right move!

snooty/language_server.py
```diff
@@ -429,7 +429,16 @@ def m_text_document__resolve(
             )
             return str(resolved_target_path)
         elif resolveType == "directive":
-            return str(self.project.config.source_path) + fileName
+            """If the filename has a .rst extension, it might be converted from
+            a YAML file. We want to get its original file path in that case."""
+            if ".rst" in fileName:
```
i80and, line 434: I think we would want `filename.endswith(".rst")`

snooty/language_server.py
```diff
@@ -429,7 +429,16 @@ def m_text_document__resolve(
...
+                filePath = str(self.project.config.source_path) + fileName
```
i80and, line 435: Python variable names should be `snake_case`. I know this file is a little inconsistent about that

I think we also want `file_path = self.project.config.source_path / fileName` so the path manipulation happens on path types rather than as string manipulation

snooty/language_server.py
```diff
@@ -429,7 +429,16 @@ def m_text_document__resolve(
...
+                fileId = self.project.get_fileid(Path(filePath))
+                finalFileId = self.project.config.source_path / fileId
```
i80and, line 437: This should be named something like `real_file_path`, since it's a path not a FileId instance

i80and (changes requested):
Nice!

Could we update the test at `test_language_server.py:test_text_doc_get_page_fileid()` to ensure this works as expected?

I think that in the test data we already load in that test, `includes/steps/migrate-compose-pr.rst` should map to `includes/steps-migrate-compose-pr.yaml`

mayaraman19:
> Could we update the test at `test_language_server.py:test_text_doc_get_page_fileid()` to ensure this works as expected?
> 
> I think that in the test data we already load in that test, `includes/steps/migrate-compose-pr.rst` should map to `includes/steps-migrate-compose-pr.yaml`

I instead updated `test_text_doc_resolve()` because the `m_text_document__get_page_fileid()` function used in `test_text_doc_get_page_fileid()` uses a different form of `get_fileid()` from the `ProjectConfig` class, which does not return the correct FileId that we are searching for in the YAML-RST conversion files. Let me know if this is not the correct way to go (I will also change the name of the `get_fileid` function I added in this PR since I can see this leading to confusion).

snooty/language_server.py
```diff
@@ -429,7 +429,18 @@ def m_text_document__resolve(
             )
             return str(resolved_target_path)
         elif resolveType == "directive":
-            return str(self.project.config.source_path) + fileName
+            """If the filename has a .rst extension, it might be converted from
+            a YAML file. We want to get its original file path in that case."""
+
+            # Strip the first slash from fileName so the / operator doesn't mess up :|
+            stripped_file_name = fileName[1:]
```
i80and, line 436: Better would be `slug.lstrip("/")`, as we do in `postprocess.clean_slug()`. Fixed offset-based string manipulation just gives me the heebie jeebies :D

i80and (changes requested):
Just one nit and then I'm good with this!

snooty/language_server.py
mayaraman19: yahoo thank you! but why are you reviewing during time off >:(

i80and (approved):

snooty/language_server.py
i80and: shame on me!! shame!!

PR #600, opened 2024-06-01
DOP-4723: Detect include loops

i80and:
### Ticket

DOP-4723

### Notes

It struck me by chance that we don't have cycle detection in our include logic. It hasn't come up, but let's make sure we handle that situation gracefully before it does.

### README updates

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -44,6 +44,7 @@ from .diagnostics import (
     ExpectedTabs,
     FetchError,
     GuideAlreadyHasChapter,
+    IncludeLoop,
     InvalidChapter,
     InvalidChild,
     InvalidContextError,
@@ -364,6 +365,13 @@ class IncludeHandler(Handler):
                         )
                     )
                 return
+
+        # Check for duplicate inclusions in this root page context.
+        if collections.Counter(fileid_stack).most_common(1)[0][1] > 1:
+            self.context.diagnostics[fileid_stack.current].append(
+                IncludeLoop(list(fileid_stack), node.span[0])
+            )
+            return
 
         include_page = self.pages.get(include_fileid)
         assert include_page is not None
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -3,7 +3,7 @@ from pathlib import Path, PurePath
 from typing import Dict, List, Optional, Sequence, Set, Tuple, Union
 
 from . import n
-from .n import SerializableType
+from .n import FileId, SerializableType
 
 
 class MakeCorrectionMixin:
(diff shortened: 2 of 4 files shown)


snooty/postprocess.py
```diff
@@ -365,6 +366,13 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
                     )
                 return
 
+        # Check for duplicate inclusions in this root page context.
```
seungpark, line 369: [minor nitpick] can we add this to the description for `bound_included_AST` as well?

snooty/postprocess.py
```diff
@@ -365,6 +366,13 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+        if collections.Counter(fileid_stack).most_common(1)[0][1] > 1:
```
seungpark, line 370: will this pick up multiple duplicate inclusions if we are only checking the first one? or would we want to include all of loops?

seungpark (commented):
overall LGTM! wanted to check if this will include multiple include loops or if that was a priority

mmeigs (commented):
Very nice. Love adding this before this nightmare actually occurs. Not approving just so Seung's question gets answered :)

rayangler (approved):

snooty/postprocess.py
i80and: Hm, do you have a concrete example of the case you're thinking of?

In the case where a page contains multiple include loops, this logic does already warn about each. We only take the `most_common(1)` because that tells us as soon as a loop starts, and prevents recursing deeper.

seungpark: ah you're right. i've tested with multiple loops within the same root file and it prints multiple warnings:

```
E             +  (<class 'snooty.diagnostics.IncludeLoop'>,
E             +   'Includes loop: index.txt -> foo.rst -> index.txt'),
E             +  (<class 'snooty.diagnostics.IncludeLoop'>,
E             +   'Includes loop: index.txt -> foo2.rst -> foo3.rst -> index.txt'),
E             +  (<class 'snooty.diagnostics.IncludeLoop'>,
E             +   'Includes loop: index.txt -> foo2.rst -> foo3.rst -> index.txt -> foo.rst -> '
E             +   'index.txt'),
E             +  (<class 'snooty.diagnostics.IncludeLoop'>,
E             +   'Includes loop: index.txt -> foo2.rst -> foo3.rst -> index.txt -> foo.rst -> '
E             +   'index.txt'),
E             +  (<class 'snooty.diagnostics.IncludeLoop'>,
E             +   'Includes loop: index.txt -> foo2.rst -> foo3.rst -> index.txt'),
E               ]
```

notably its still sending the full fileid_stack!

seungpark (approved):
LGTM !

snooty/postprocess.py
seungpark: perhaps we can include a multi loop? not a big deal, it was my oversight!

PR #601, opened 2024-06-05
Dop 4623

mayaraman19:
### Ticket

DOP-4623

### Notes

Ugh, there's a lot going on in my brain for this but hopefully I get everything. 

- in addition to Snooty's changes, this adds support for getting card icons from their names in the list [here](https://mongodbcom.website.prod.corp.mongodb.com/cms/icons) rather than having to store them in the content repo. This makes it easier to get images in their dark mode version, since we don't have to store both versions and can instead access them via a URL.
- If, however, we can't access the necessary image from the above list, the `icon-dark` option has been added to cards, as well as a warning if `icon` comes from an image stored in the content repo but `icon-dark` is not given.

Example parser output now looks like this:
image

Relevant Snooty PR: <WILL BE LINKED HERE LATER>

### README updates

### DIFF
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -850,6 +850,20 @@ class IconMustBeDefined(Diagnostic):
             end,
         )
 
+class CardIconString(Diagnostic):
+    severity = Diagnostic.Level.warning
+
+    def __init__(
+        self, 
+        image_argument: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None
+    ) -> None:
+        super().__init__(
+            f"""Card icons are migrating to strings instead of path names. If {image_argument} can be a string, use that instead.""",
+            start, 
+            end,
+        )
 
 class InvalidOpenApiResponse(Diagnostic):
     severity = Diagnostic.Level.error
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -58,6 +58,7 @@ from .diagnostics import (
     ExpectedStringArg,
     FetchError,
     IconMustBeDefined,
+    CardIconString,
     ImageSuggested,
     InvalidDirectiveStructure,
     InvalidField,
@@ -1046,8 +1047,16 @@ class JSONVisitor:
             image_argument = options.get("icon")
             url_argument = options.get("url")
 
-            if image_argument:
+            # if the image is a path, we need a dark mode version as well
+            if image_argument and "/" in image_argument:
+                self.diagnostics.append(CardIconString(image_argument, line))
                 self.validate_and_add_asset(doc, image_argument, line)
+                dark_mode_image_argument = options.get("icon-dark")
+                if dark_mode_image_argument:
+                    self.validate_and_add_asset(doc, dark_mode_image_argument, line)
+                else:
+                    self.diagnostics.append(ExpectedOption(name, "icon-dark", line))
+
             if url_argument and not url_argument.startswith("http"):
                 self.validate_relative_url(url_argument, line)
 
@@ -1063,6 +1072,7 @@ class JSONVisitor:
             static_asset = self.add_static_asset(image_argument, upload=True)
             self.pending.append(PendingFigure(doc, static_asset, self.dependencies))
         except OSError as err:
+            # if it has a slash is a path - 
             self.diagnostics.append(
                 CannotOpenFile(Path(image_argument), err.strerror, line)
             )
--- a/snooty/rstparser.py
+++ b/snooty/rstparser.py
@@ -117,33 +117,6 @@ class LegacyTabsDefinition(nodes.Node):
     hidden: Optional[bool]
     tabs: List[LegacyTabDefinition]
 
-
-@checked
-@dataclass
-class CardDefinition(nodes.Node):
-    """Represents a Card within a CardGroup.
-
-    Attributes:
-        id          Unique identifier for the card, to be used as an anchor tag
-        headline    Card title heading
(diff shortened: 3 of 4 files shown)


snooty/rstparser.py
```diff
@@ -118,32 +118,6 @@ class LegacyTabsDefinition(nodes.Node):
     tabs: List[LegacyTabDefinition]
 
 
-@checked
```
i80and, line 121: all this removal brings me great joy!

snooty/diagnostics.py
```diff
@@ -851,6 +851,22 @@ def __init__(
...
+    severity = Diagnostic.Level.warning
+
+    def __init__(
+        self,
+        image_argument: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ) -> None:
+        super().__init__(
+            f"""Card icons are migrating to strings instead of path names. If {image_argument} can be a string, use that instead.""",
```
rayangler, line 864: (Non-blocking) Would it be worth including more information on what the "string" is supposed to be or how to know what the string is supposed to be?

Also, I know that we have a strategy for handling marketing's icons through the web assets URL, but do we have a strategy for other icons like for drivers/languages/[integrations](https://www.mongodb.com/docs/languages/java/#integrations--community-maintained-)? If we don't currently support all existing icons by using strings, I'm afraid of causing noise and confusion through this warning for card icons that can't immediately be updated. Should we consider adding this warning at a later time once we fully support everything?

snooty/parser.py
```diff
@@ -1046,11 +1047,23 @@ def _locate_text(text: str) -> int:
...
+                    self.validate_and_add_asset(doc, image_argument, line)
+                    dark_mode_image_argument = options.get("icon-dark")
+                    if dark_mode_image_argument:
+                        self.validate_and_add_asset(doc, dark_mode_image_argument, line)
+                    else:
+                        self.diagnostics.append(ExpectedOption(name, "icon-dark", line))
+            else:
+                if image_argument:
+                    self.validate_and_add_asset(doc, image_argument, line)
+
```
mmeigs, line 1066: total nit: you could consolidate a tiny bit more if you wanted, I believe:
```
if key == "mongodb:card" and image_argument:
  self.validate_and_add_asset(doc, image_argument, line)
  if "/" in image_argument:
    ... the rest of that block of code (but with the validate_and_add_asset line unneeded now
```

snooty/rstparser.py
mmeigs: yes, pretty red :)

snooty/parser.py
```diff
@@ -1046,11 +1047,23 @@ def _locate_text(text: str) -> int:
...
 
-            if image_argument:
-                self.validate_and_add_asset(doc, image_argument, line)
             if url_argument and not url_argument.startswith("http"):
                 self.validate_relative_url(url_argument, line)
 
+            # for cards - if the image is a path, we need a dark mode version as well
+            if key == "mongodb:card":
+                if image_argument and "/" in image_argument:
+                    self.diagnostics.append(CardIconString(image_argument, line))
```
seungpark, line 1056: similar to @rayangler s comment above but wouldn't [this PR](https://github.com/mongodb/docs-ecosystem/pull/972/files) still have a warning ? 

think we should allow name and path, but require a `icon-dark` if its a path

seungpark (changes requested):

snooty/diagnostics.py
mayaraman19: that sounds good!

snooty/parser.py
mayaraman19: i do want to consolidate but i think the `validate_and_add_asset()` function should only be called on path arguments (i.e. with arguments with "/") :(

snooty/diagnostics.py
mayaraman19: i am wondering if the warning that a dark mode icon hasn't been included should be silenced until we are closer to releasing dark mode?

snooty/parser.py
i80and: Correct -- `validate_and_add_asset()` flags a path to be read down the road -- it gets turned into a concrete Path down the call chain (this is a little bit of a code smell).

snooty/diagnostics.py
rayangler: I think I'm okay with waiting if we need the time to sort out all of the dark mode icons for cards

snooty/parser.py
mmeigs: aha, I didn't see the difference in the arguments! Thanks!

snooty/diagnostics.py
mayaraman19: Removed the warning!

rayangler (approved):

snooty/parser.py
```diff
@@ -1046,11 +1046,20 @@ def _locate_text(text: str) -> int:
...
                 self.validate_relative_url(url_argument, line)
 
+            # for cards - if the image is a path, we need a dark mode version as well
+            if key == "mongodb:card":
+                if image_argument and "/" in image_argument:
+                    self.validate_and_add_asset(doc, image_argument, line)
+                    dark_mode_image_argument = options.get("icon-dark")
+                    if dark_mode_image_argument:
+                        self.validate_and_add_asset(doc, dark_mode_image_argument, line)
+            else:
```
i80and, line 1059: nit: Collapse this into a single `elif image argument:`

i80and (changes requested):

i80and (approved):

PR #611, opened 2024-07-16
DOP-4690: prepare parser for new collapsible directive

seungpark:
### Ticket

DOP-4690

### Notes

**Changes from this PR**
- Adds `.. collapsible:: ` directive into our rstspec, and handles new ReSTructured text into expected AST
has `heading:string (required)` and `sub_heading.string` options, along with a required `content_type:block` 
- Adds Error level Diagnostics for missing content block within collapsible directive
- Adds Error level Diagnostics for nested collapsibles
- Inserts the `heading` option from new directive into `page.ast.options.headings[]` during postprocess so that this can be included in the right column for `On this page`. Discussed offline with @i80and that the alternative to this would be to enforce a heading and subheading structure within the `content_type`, but no strong opinions here

### README updates

### DIFF
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -944,3 +944,19 @@ class OrphanedPage(Diagnostic):
             0,
             None,
         )
+
+
+class NestedDirective(Diagnostic):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        nested_directive: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ):
+        super().__init__(
+            f"""Nested directive detected: {nested_directive}.""",
+            start,
+            end,
+        )
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -956,7 +956,7 @@ class NestedDirective(Diagnostic):
         end: Union[None, int, Tuple[int, int]] = None,
     ):
         super().__init__(
-            f"""Nested directive detected: {nested_directive}.""",
+            f"""Nesting of {nested_directive} directives prohibited""",
             start,
             end,
         )
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -1059,6 +1059,12 @@ class JSONVisitor:
             elif image_argument:
                 self.validate_and_add_asset(doc, image_argument, line)
 
+        elif key in {"mongodb:collapsible"}:
+            if not node.children or len(node.children) == 0:
+                self.diagnostics.append(
+                    MissingChild("mongodb:card", "content block", line)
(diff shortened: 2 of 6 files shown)


snooty/parser.py
```diff
@@ -1059,6 +1059,12 @@ def _locate_text(text: str) -> int:
             elif image_argument:
                 self.validate_and_add_asset(doc, image_argument, line)
 
+        elif key in {"mongodb:collapsible"}:
```
i80and, line 1062: this should just be `key == "mongodb:collapsible"`, otherwise we're creating a set object just for a __contains__ call

snooty/parser.py
```diff
@@ -1059,6 +1059,12 @@ def _locate_text(text: str) -> int:
...
+            if not node.children or len(node.children) == 0:
```
i80and, line 1063: I believe this can just be `if not node.children`?

i80and (changes requested):
Just two trivial nits!

snooty/parser.py
seungpark: good catch. updating

seungpark:
> Just two trivial nits!

sorry for the oversight but from [this thread](https://www.figma.com/design/7cR6eWtFvKTnJIPL8ZsTaB?node-id=4807-6661#850607854) it looks like we want to catch nested collapsibles as well! added a postprocess and a test case for this PTAL

snooty/diagnostics.py
```diff
@@ -944,3 +944,19 @@ def __init__(self) -> None:
...
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        nested_directive: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ):
+        super().__init__(
+            f"""Nested directive detected: {nested_directive}.""",
```
i80and, line 959: Could we make this error message more descriptive for writers? Perhaps along the lines of f"""Nesting of "{nested_directive}" directives prohibited"""

snooty/postprocess.py
```diff
@@ -1829,6 +1844,27 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
             self.current_img_index += 1
 
 
+class CollapsibleHandler(Handler):
+    """ """
```
i80and, line 1848: Could we fill out this docstring?

i80and (changes requested):
Two more nits in the postprocessor bit!

snooty/diagnostics.py
seungpark: i like the verbosity! updated

i80and (approved):

PR #613, opened 2024-08-05
DOP-4692: Implement wayfinding directives

rayangler:
### Ticket

DOP-4692
[Related frontend PR](https://github.com/mongodb/snooty/pull/1192)
[Designs](https://www.figma.com/design/7cR6eWtFvKTnJIPL8ZsTaB/Content-Template-Projects?node-id=3147-3388&t=PnibGktO5uVkJEtU-0)
[rST example](https://github.com/rayangler/cloud-docs/blob/test-wayfinding/source/testing-wayfinding.txt)

### Notes

* I opted to create a separate directive for the description inside of the `wayfinding` directive instead of using some `description` option because (1) it allows the component to be more flexible for different use cases and copy if needed, and (2) it allows the use of an `include` directive to make it easier to share copy for related instances of the directive.

### README updates

### DIFF
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -968,12 +968,12 @@ class UnknownWayfindingOption(Diagnostic):
     def __init__(
         self,
         invalid_id: str,
-        valid_ids: AbstractSet[str],
+        valid_ids: List[str],
         start: Union[int, Tuple[int, int]],
         end: Union[None, int, Tuple[int, int]] = None,
     ):
         super().__init__(
-            f"Wayfinding option id {invalid_id} is not valid. Expected one of the following: {sorted(valid_ids)}",
+            f"Wayfinding option id {invalid_id} is not valid. Expected one of the following: {valid_ids}",
             start,
             end,
         )
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -960,3 +960,19 @@ class NestedDirective(Diagnostic):
             start,
             end,
         )
+
+
+class UnknownWayfindingOption(Diagnostic):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        invalid_id: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ):
+        super().__init__(
+            f"Wayfinding option id {invalid_id} is not valid.",
+            start,
+            end,
+        )
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -968,11 +968,12 @@ class UnknownWayfindingOption(Diagnostic):
     def __init__(
         self,
         invalid_id: str,
+        valid_ids: AbstractSet[str],
         start: Union[int, Tuple[int, int]],
         end: Union[None, int, Tuple[int, int]] = None,
     ):
         super().__init__(
-            f"Wayfinding option id {invalid_id} is not valid.",
+            f"Wayfinding option id {invalid_id} is not valid. Expected one of the following: {sorted(valid_ids)}",
             start,
             end,
         )
--- a/snooty/diagnostics.py
+++ b/snooty/diagnostics.py
@@ -976,3 +976,19 @@ class UnknownWayfindingOption(Diagnostic):
             start,
             end,
         )
+
+
+class DuplicateWayfindingOption(Diagnostic):
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        invalid_id: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ):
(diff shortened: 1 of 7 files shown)


snooty/parser.py
```diff
@@ -644,6 +650,99 @@ def handle_tabset(self, node: n.Directive) -> None:
...
+                )
+                continue
+
+            option_details = expected_options_dict[option_id]
+            child.options["title"] = option_details.title
+            child.options["language"] = option_details.language
+            valid_opts.append(child)
+            used_ids.add(option_id)
+
+        def sort_key(node: n.Node) -> tuple[bool, str, str]:
```
seungpark, line 718: [minor] can we take the node arg to be type n.Directive

snooty/parser.py
```diff
@@ -644,6 +650,99 @@ def handle_tabset(self, node: n.Directive) -> None:
...
+            assert isinstance(node, n.Directive)
+            # Associate the child node with the actual wayfinding option
+            wayfinding_option = expected_options_dict[node.options["id"]]
+            return (
+                not wayfinding_option.show_first,
+                wayfinding_option.language,
+                wayfinding_option.title,
+            )
+
+        new_children = sorted(valid_opts, key=sort_key)
```
seungpark, line 729: minor. `new_children` is typed as `any` here. can we do a `valid_opts.sort()` ?

i80and: `valid_opts.sort()` should be safe -- it's a behavioral change, but the only subsequent use of `valid_opts` is checking its size, so it's all good. Plus, in-place sort

snooty/parser.py
rayangler: Sure. I originally made it `n.Node` because `node.children` wasn't compatible with `List[n.Directive]`, which was causing typing issues with the list of nodes/directives here. Casting the list before reassignment seems to do the trick though. Let me know if it looks good to you

snooty/parser.py
rayangler: Whoops, thanks for catching. Made the change

snooty/test_parser.py
```diff
@@ -3874,3 +3877,252 @@ def test_collapsible() -> None:
...
+   .. wayfinding-option:: https://www.mongodb.com/docs/
+      :id: c
+
+   .. wayfinding-option:: https://www.mongodb.com/docs/
+      :id: java-sync
+    
+   .. wayfinding-option:: https://www.mongodb.com/docs/
+      :id: bad-id
+   
+   .. wayfinding-option:: https://www.mongodb.com/docs/
```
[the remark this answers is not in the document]
seungpark: ah my overlook! i thought this was a missing argument to the actual `.. wayfinding::` component 🤦

snooty/parser.py
```diff
@@ -644,6 +650,99 @@ def handle_tabset(self, node: n.Directive) -> None:
...
+            # Type is ambiguous now despite if statements above
+            assert isinstance(child, n.Directive)
+            if child.name == expected_child_desc_name:
+                valid_desc = child
+                continue
+
+            option_id = child.options.get("id")
+
+            if not (child.argument and option_id):
+                # Don't append diagnostic since docutils should already
```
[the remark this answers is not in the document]
seungpark: see comment above! my overlook. this is fine with me 👍

snooty/parser.py
```diff
@@ -644,6 +650,99 @@ def handle_tabset(self, node: n.Directive) -> None:
...
+            if invalid_child:
+                self.diagnostics.append(
+                    InvalidChild(
+                        invalid_child,
+                        wayfinding_name,
+                        f"{expected_child_opt_name} or {expected_child_desc_name}",
+                        child_line_start,
+                    )
+                )
+                continue
```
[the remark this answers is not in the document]
seungpark: i guess its a matter of `do we catch all errors in one go and not have to go through different errors every time` vs `catch early and return early`. (since this is the parser, and not post process) i'm leaning towards the former, but with all the CI integration time costs, i'm fine with catching more errors in the go as well.

seungpark (approved):
this LGTM ! TDD ftw

snooty/diagnostics.py
```diff
@@ -960,3 +960,35 @@ def __init__(
...
+    severity = Diagnostic.Level.error
+
+    def __init__(
+        self,
+        invalid_id: str,
+        start: Union[int, Tuple[int, int]],
+        end: Union[None, int, Tuple[int, int]] = None,
+    ):
+        super().__init__(
+            f"Wayfinding option id {invalid_id} is not valid.",
```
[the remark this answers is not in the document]
rayangler: I updated it to include the full list of available IDs. I'm wondering if it's better to include a list of all option ids like right now, or if it'd be better to wait until the end of the for loop to list out which ids are leftover for the user to pick from. What do you think?

i80and (approved):

PR #614, opened 2024-08-27
DOP-4685: Implement method-selector

rayangler:
### Ticket

DOP-4685
[Example rST](https://github.com/rayangler/cloud-docs/blob/ace163b51c2f157f38a6d044d2a5faea63b287d3/source/testing-method-selector.txt)
[Designs](https://www.figma.com/design/7cR6eWtFvKTnJIPL8ZsTaB/Content-Template-Projects?node-id=3473-22389&t=RNnOtvg64NtvtUHi-0)
[Related frontend PR](https://github.com/mongodb/snooty/pull/1208)

### Notes

* There were a lot of similarities to the `wayfinding` directive, so I tried my best to make the logic more reusable. Let me know if there are any issues with anything.
* There's a weird bug where headings inside of this directive aren't captured in the `ContentsHandler` for some reason. I'm hoping to look into this as a separate ticket since it doesn't affect the component itself. I don't think this should be a blocker for this PR considering the "On This Page" component on the frontend will have the same issue with showing headings from unselected tabs, which is a less than ideal UX either way.

### README updates

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -743,8 +743,73 @@ class JSONVisitor:
 
         node.children = cast(List[n.Node], valid_children)
 
-    def handle_method_selector(self, node: n.Directive):
+    def is_valid_child(self, parent: n.Node, child: MutableSequence[n.Node], expected_children_names: Set[str], directives_only: bool = False):
+        # TODO: Fill this out and reuse for wayfinding, method-selector, and method-option
         pass
+
+    def handle_method_selector(self, node: n.Directive):
+        expected_options = specparser.Spec.get().method_selector["options"]
+        expected_options_dict = {option.id: option for option in expected_options}
+        expected_child_name = "method-option"
+        
+        valid_children: List[n.Directive] = []
+        used_ids: Set[str] = set()
+
+        # Validate children
+        for child in node.children:
+            child_line_start = child.start[0]
+
+            invalid_child = None
+            if not isinstance(child, n.Directive):
+                # Catches additional unwanted types like Paragraph
+                invalid_child = child.type
+            elif child.name != expected_child_name:
+                invalid_child = child.name
+
+            if invalid_child:
+                self.diagnostics.append(
+                    InvalidChild(
+                        invalid_child,
+                        node.name,
+                        expected_child_name,
+                        child_line_start,
+                    )
+                )
+                continue
+
+            option_id = child.options.get("id")
+            if not option_id:
+                # Don't append diagnostic since docutils should already
+                # complain about missing argument and ID option
+                continue
+
+            if not option_id in expected_options_dict:
+                available_ids = list(expected_options_dict.keys())
+                available_ids.sort()
+                # TODO: Generalize UnknownOption?
+                self.diagnostics.append(
+                    UnknownWayfindingOption(option_id, available_ids, child_line_start)
+                )
+                continue
+
+            if option_id in used_ids:
+                # TODO: Generalize DuplicateOption?
+                self.diagnostics.append(
+                    DuplicateWayfindingOption(option_id, child_line_start)
+                )
+                continue
+
+            # The Drivers option should be encouraged to be first
+            if option_id == "drivers" and valid_children:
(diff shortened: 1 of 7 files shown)


i80and (changes requested):
Just a couple minor requests!

snooty/parser.py
```diff
@@ -664,53 +678,17 @@ def handle_wayfinding(self, node: n.Directive) -> None:
...
-                )
-                continue
-
-            if option_id in used_ids:
-                self.diagnostics.append(
-                    DuplicateWayfindingOption(option_id, child_line_start)
-                )
+            try:
+                self.check_valid_child(node, child, expected_children_names)
+                assert isinstance(child, n.Directive)
```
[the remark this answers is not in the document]
rayangler: Done

snooty/parser.py
```diff
@@ -743,6 +721,124 @@ def sort_key(node: n.Directive) -> tuple[bool, str, str]:
 
         node.children = cast(List[n.Node], valid_children)
 
+    def check_valid_child(
+        self,
+        parent: n.Directive,
+        child: n.Node,
+        expected_children_names: Set[str],
+    ) -> None:
```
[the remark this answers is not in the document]
rayangler: Done

snooty/parser.py
```diff
@@ -560,6 +568,12 @@ def dispatch_departure(self, node: tinydocutils.nodes.Node) -> None:
         elif isinstance(popped, n.Directive) and popped.name == "wayfinding":
             self.handle_wayfinding(popped)
 
+        elif isinstance(popped, n.Directive) and popped.name == "method-selector":
```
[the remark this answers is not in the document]
rayangler: Makes sense. Happy to factor out an `if not isinstance(popped, n.Directive)` if we want to bundle it as part of this PR. Interested to see what a refactor of this would look like 👀

snooty/parser.py
```diff
@@ -107,6 +109,12 @@ class ProjectLoadError(Exception):
     pass
 
 
+class ChildValidationError(Exception):
+    """Empty exception expected to continue"""
```
mmeigs, line 113: For me... what does this mean? Not sure if this isn't clear or if I'm misunderstanding.

snooty/postprocess.py
```diff
@@ -1897,17 +1903,75 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
-        if node.name != self.wayfinding_name:
+        if node.name != self.directive_name:
             self.nesting_level -= 1
 
-    def enter_page(self, fileid_stack: FileIdStack, page: Page) -> None:
-        self.nesting_level = 0
-        self.wayfinding_detected = False
+
+class WayfindingHandler(NestedDirectiveHandler):
+    """Handles page-level validations for method-selector directive and its children."""
```
mmeigs, line 1916: Should this be "wayfinding directive"?

i80and (approved):

mmeigs (approved):
Nice!

snooty/parser.py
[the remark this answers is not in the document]
i80and: My thinking is instead of God's Ugliest elif chain this should be a list of handlers, ideally each in their own file. Closer to what we do in the postprocessor.

snooty/parser.py
rayangler: Nah, you're right. It can be quite unclear without context. I tried to convey that this exception intentionally doesn't do much given that it's only supposed to be used to help call "continue" in some loops used for validation when it's raised.

I went ahead and just removed the docstring because it sounds unnecessarily and confusingly prescriptive. The class name and current usage are probably good enough to give an idea of what it's for. Wdyt?

snooty/postprocess.py
rayangler: Yes! Thank you for catching. Fixed

snooty/parser.py
mmeigs: I'm fine with this, thanks!

PR #616, opened 2024-09-09
DOP-4983: controlled sections and refs for collapsibles

seungpark:
### Ticket

DOP-4983

### Notes
- This PR updates sectioning logic for collapsibles: the `heading` option will be at the same level as the sibling content above, and content within collapsibles should be considered another level down.
- Also addressed the third point in the issue, where refs that were pointing to collapsibles were not pulling the content from the heading option.

### README updates

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -1568,14 +1568,25 @@ class AddTitlesToLabelTargetsHandler(Handler):
         self.pending_targets: List[n.Node] = []
 
     def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
-        if not isinstance(node, (n.Target, n.Section, n.TargetIdentifier)):
+        if not isinstance(node, (n.Target, n.Section, n.TargetIdentifier)) and not (
+            isinstance(node, n.Directive) and "heading" in node.options.keys()
+        ):
             self.pending_targets = []
 
         if isinstance(node, n.Target) and node.domain == "std" and node.name == "label":
             self.pending_targets.extend(node.children)
-        elif isinstance(node, n.Section):
+        elif isinstance(node, n.Section) or (
+            isinstance(node, n.Directive) and "heading" in node.options.keys()
+        ):
             for target in self.pending_targets:
-                heading = next(node.get_child_of_type(n.Heading), None)
+                if isinstance(node, n.Section):
+                    heading = next(node.get_child_of_type(n.Heading), None)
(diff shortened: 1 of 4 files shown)


snooty/postprocess.py
```diff
@@ -1568,14 +1568,27 @@ def __init__(self, context: Context) -> None:
...
             self.pending_targets.extend(node.children)
-        elif isinstance(node, n.Section):
+        elif isinstance(node, n.Section) or (
+            isinstance(node, n.Directive) and "heading" in node.options.keys()
+        ):
             for target in self.pending_targets:
-                heading = next(node.get_child_of_type(n.Heading), None)
+                if isinstance(node, n.Section):
+                    heading = next(node.get_child_of_type(n.Heading), None)
+                if isinstance(node, n.Directive) and "heading" in node.options.keys():
```
i80and, line 1584: `elif`, and you can just write `"heading" in node.options`: the `__in__` dunder method on dict implies looking at keys

snooty/postprocess.py
```diff
@@ -1568,14 +1568,27 @@ def __init__(self, context: Context) -> None:
         self.pending_targets: List[n.Node] = []
 
     def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
-        if not isinstance(node, (n.Target, n.Section, n.TargetIdentifier)):
+        if not isinstance(node, (n.Target, n.Section, n.TargetIdentifier)) and not (
+            isinstance(node, n.Directive) and "heading" in node.options.keys()
```
i80and, line 1572: You don't need to specify the `keys()` method

i80and (changes requested):
Trivial nits

snooty/postprocess.py
seungpark: done! thanks for the 101 lesson ☺️ removed all the calls to keys

snooty/postprocess.py
seungpark: removed!

seungpark:
hey @i80and ! i think i may have overlooked somethings in this PR:
1. [this iterator](https://github.com/mongodb/snooty-parser/pull/616/files#diff-5e66e8539da7ed36bf30c7a27807d04eb3c8c9a0d78649afcdc4a90cb1871efeR576) will always be true, have to check that its not empty
2. A new section should be added within collapsible content (if there are none) as described above. but also, a collapsible itself should be in a section, as its heading should always be h2 and subsequent headers should follow suit (h3, subtitle) afterwards

More fixes coming up

PR #618, opened 2024-09-12
DOP-4983: improve heading content and sectioning

seungpark:
### Ticket

DOP-4983

### Notes
- Ticket fixes sectioning depth by adding a section into collapsible content (if one does not already exist in the direct children)
- Ticket addresses heading options for the AST, where wrong headings would be contained within the page options
- Unit tests have been added to identify above situations

### README updates

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -308,9 +308,6 @@ class JSONVisitor:
             )
         elif isinstance(node, rstparser.directive):
             directive = self.handle_directive(node, line)
-            # if directive and node["name"] == "collapsible":
-            #     self.state.append(n.Section((node.get_line(),), [directive]))
-            # elif directive:
             if directive:
                 self.state.append(directive)
         elif isinstance(node, tinydocutils.nodes.Text):
@@ -576,6 +573,8 @@ class JSONVisitor:
             self.handle_method_option(popped)
 
         elif isinstance(popped, n.Directive) and popped.name == "collapsible":
+            html5_id = util.make_html5_id(popped.options.get("heading", "")).lower()
+            popped.options["id"] = html5_id
             has_section = any(True for _ in popped.get_child_of_type(n.Section))
             has_heading = any(True for _ in popped.get_child_of_type(n.Heading))
             if not has_section and not has_heading:
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -575,10 +575,7 @@ class JSONVisitor:
         elif isinstance(popped, n.Directive) and popped.name == "collapsible":
             html5_id = util.make_html5_id(popped.options.get("heading", "")).lower()
             popped.options["id"] = html5_id
-            has_section = any(True for _ in popped.get_child_of_type(n.Section))
-            has_heading = any(True for _ in popped.get_child_of_type(n.Heading))
-            if not has_section and not has_heading:
-                popped.children = [n.Section((node.get_line(),), popped.children)]
+            popped.children = [n.Section((node.get_line(),), popped.children)]
 
     def handle_facet(self, node: rstparser.directive, line: int) -> None:
         if "values" not in node["options"] or "name" not in node["options"]:
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -308,6 +308,9 @@ class JSONVisitor:
             )
         elif isinstance(node, rstparser.directive):
             directive = self.handle_directive(node, line)
+            # if directive and node["name"] == "collapsible":
+            #     self.state.append(n.Section((node.get_line(),), [directive]))
+            # elif directive:
             if directive:
                 self.state.append(directive)
         elif isinstance(node, tinydocutils.nodes.Text):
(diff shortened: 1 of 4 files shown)


snooty/parser.py
```diff
@@ -573,9 +573,11 @@ def dispatch_departure(self, node: tinydocutils.nodes.Node) -> None:
             self.handle_method_option(popped)
 
         elif isinstance(popped, n.Directive) and popped.name == "collapsible":
-            section = popped.get_child_of_type(n.Section)
-            heading = popped.get_child_of_type(n.Heading)
-            if not section and not heading:
+            html5_id = util.make_html5_id(popped.options.get("heading", "")).lower()
+            popped.options["id"] = html5_id
```
rayangler, line 577: Will we ever need to ensure uniqueness in collapsible heading IDs like we do with headings in the postprocessor's `HeadingHandler`? I'm thinking of a case where similar collapsible headings are used in method selectors or tabs.

snooty/parser.py
```diff
@@ -573,9 +573,11 @@ def dispatch_departure(self, node: tinydocutils.nodes.Node) -> None:
...
+            has_section = any(True for _ in popped.get_child_of_type(n.Section))
+            has_heading = any(True for _ in popped.get_child_of_type(n.Heading))
+            if not has_section and not has_heading:
```
rayangler, line 580: Would it make sense to create a section regardless of the contents of the collapsible? Seems like we're always increasing the section depth in the `ContentsHandler` when a `collapsible` is entered anyways.

The [updated staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOCSP-43269-2/cloud-docs/seung.park/DOP-4983-frontend/tutorial/insert-data-cli/#required-access) still shows an h2 nested under the last collapsible. My understanding is that "Required Access" should be an h3, similar to the headings of the steps in the second collapsible.

rayangler: I think two sections *should* be fine. The structure would be somewhat similar to if an h2 heading immediately wraps an h3 heading. The main difference is the order of the section (section -> h2 -> section -> h3 vs. collapsible -> section -> section -> h3). Wrapping the collapsible in a section like you suggested should help resolve the issue while keeping order consistent, so I think that's also a good idea

snooty/parser.py
seungpark: done! added logic to ensure uniqueness for heading options in collapsible contents. PTAL and lmk if that works !

snooty/parser.py
seungpark: I've also updated that above staging link to be parsed from this branch!

snooty/postprocess.py
```diff
@@ -527,18 +527,21 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
 
         if isinstance(node, n.Directive) and node.name == "collapsible":
-            html5_id = util.make_html5_id(node.options["heading"]).lower()
-            node.options["id"] = html5_id
+            self.current_depth += 1
+            if self.current_depth - 1 > self.contents_depth:
+                return
             self.headings.append(
                 ContentsHandler.HeadingData(
                     self.current_depth,
```
rayangler, line 535: Looks like we'll have a double depth increment here due to the section being added. This will mean that for an example like:

```
.. contents::
    :depth: 1

===================
Heading of the page
===================

.. collapsible::
    :heading: Collapsible heading
    :sub_heading: Subheading

    ~~~~~~~~~~~~~~~
    This is content
    ~~~~~~~~~~~~~~~
```

the user will have to bump `:depth:` to 3 instead of 2 if they want to capture `This is content` in the "On This Page" component.

We can probably just rely on the presence of the section to handle incrementing, while artificially adding 1 when adding the collapsible heading to account for the section being after the directive. What do you think?

```py
if isinstance(node, n.Directive) and node.name == "collapsible":
    self.headings.append(
        ContentsHandler.HeadingData(
            # Add 1 since section appears as a child
            self.current_depth + 1,
            node.options["id"],
            [n.Text(node.span, node.options["heading"])],
        )
    )
```

snooty/postprocess.py
```diff
@@ -527,18 +527,21 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+                    node.options["id"],
                     [n.Text(node.span, node.options["heading"])],
                 )
             )
 
     def exit_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
-        if isinstance(node, n.Section):
+        if isinstance(node, n.Section) or (
+            isinstance(node, n.Directive) and node.name == "collapsible"
+        ):
```
rayangler, line 544: Likewise, we can probably let the presence of a section handle changing depth

snooty/parser.py
rayangler: I think it looks good! Thanks!

rayangler (commented):
Just 1 more thing I noticed about section depth, but other looks great!

snooty/postprocess.py
seungpark: genius! done! added above test case 👍

snooty/test_postprocess.py
```diff
@@ -3484,6 +3484,30 @@ def test_collapsible_headings() -> None:
...
+
+    ~~~~~~~~~~~~~~~
+    This is content
+    ~~~~~~~~~~~~~~~
+""",
+        }
+    ) as result:
+        page = result.pages[FileId("index.txt")]
+
+    with make_test(
```
rayangler, line 3499: Are we testing anything for this case? Should we remove? Otherwise, we can probably consolidate with this other [test case](https://github.com/mongodb/snooty-parser/pull/618/commits/6718998a14c15754262af988d3a20b520f064f3b#diff-bfac8a18bf662c13854408d2116c9293dc488fde36361670d1c7af27becdafa3R3621)

seungpark: removed this null test!

PR #619, opened 2024-09-12
DOP-4917: Add video options for video metadata

rayangler:
### Ticket

DOP-4917
[Related frontend PR](https://github.com/mongodb/snooty/pull/1236)
[Example rST](https://github.com/rayangler/cloud-docs/commit/93ab9b1cc34c1a461e8266da44f07687acc49c49)

### Notes

* I know the original ticket didn't ask for a `description` option, but I saw this [Google Rich Result test](https://search.google.com/test/rich-results/result?id=NWG835UY5-cwJQ_1l8KR-A) that showed a single warning and figured it's worth including for completion.
* All options are not required for the `video` directive to be used on a page, in order to allow for backwards compatibility. However, a warning will be added whenever one of the options is used without the use of all other options for structured data. The frontend will handle whether to show the structured data or not based on the options available.
* This PR should be good to merge without the need for the frontend PR as the existence of the new options shouldn't break anything

### README updates

### DIFF
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -1232,9 +1232,14 @@ class JSONVisitor:
                 self.validate_and_add_asset(doc, icon_argument, line)
 
         elif name == "video":
-            sd_req_option_names = ["title", "thumbnail-url", "upload-date", "description"]
+            sd_req_option_names = [
+                "title",
+                "thumbnail-url",
+                "upload-date",
+                "description",
+            ]
             missing_option_names = []
-            
+
             for option_name in sd_req_option_names:
                 val = options.get(option_name, None)
                 if not val:
@@ -1242,8 +1247,12 @@ class JSONVisitor:
 
             # We want to encourage defining all of these options together or not at all for structured data SEO
             missing_options_len = len(missing_option_names)
-            if (missing_options_len > 0 and missing_options_len != len(sd_req_option_names)):
-                self.diagnostics.append(MissingStructuredDataFields(name, missing_option_names, line))
+            if missing_options_len > 0 and missing_options_len != len(
+                sd_req_option_names
+            ):
+                self.diagnostics.append(
+                    MissingStructuredDataFields(name, missing_option_names, line)
+                )
 
         elif name in {"pubdate", "updated-date"}:
             if "date" in node:
--- a/snooty/parser.py
+++ b/snooty/parser.py
@@ -71,6 +71,7 @@ from .diagnostics import (
     MissingAssociatedToc,
     MissingChild,
     MissingFacet,
+    MissingStructuredDataFields,
     RemovedLiteralBlockSyntax,
     TabMustBeDirective,
     TodoInfo,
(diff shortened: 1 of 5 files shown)


snooty/rstspec.toml
```diff
@@ -886,7 +886,14 @@ options.type = "string"
...
 help = "Embed a video in your article"
 argument_type = "uri"
-example = """.. video:: ${1: full video url}"""
+options.title = "string"
+options.description = "string"
+options.thumbnail-url = "uri"
+options.upload-date = "iso_8601"
+example = """.. video:: ${1: full video url}
+   :thumbnail-url: ${2: video's thumbnail url}
+   :upload-date: ${3: video's original upload date in ISO-8601 format}
```
seungpark, line 895: minor. can we provide examples for iso 8601 for posterity

snooty/rstspec.toml
```diff
@@ -886,7 +886,14 @@ options.type = "string"
 [directive.video]
 help = "Embed a video in your article"
 argument_type = "uri"
-example = """.. video:: ${1: full video url}"""
+options.title = "string"
```
seungpark, line 889: do we need to warn for partial options. in the case that they put only title and [wondering why SD does not show up](https://github.com/mongodb/snooty/pull/1236/files#diff-3f8a5baf91118745fc1b50041ac882a0ce473b5aab68c615832d9aad05a0ac5bR77)

seungpark (commented):
minor comments below! wdyt about the optional fields being bundle-required?

snooty/rstspec.toml
rayangler: Good idea. I reorganized the example a bit to provide more concrete examples/placeholders. Let me know what you think

snooty/rstspec.toml
rayangler: I originally wanted to keep usage of the options more loose in case we wanted to use them for non-structured data things in the future. That being said, we're only using them for structured data for now, so I went ahead and added some logic to warn against this.

Just noting here that even though `description` is optional for Google, I made it "required" in the parser to encourage usage of it along with the other options and to simplify the logic. Let me know if this works for you or if you need me to change anything.

snooty/parser.py
```diff
@@ -1230,6 +1231,29 @@ def _locate_text(text: str) -> int:
             if icon_argument:
                 self.validate_and_add_asset(doc, icon_argument, line)
 
+        elif name == "video":
+            sd_req_option_names = [
```
seungpark, line 1235: [nit] is it possible to get this from the defined options vs defining them twice? im fine with this as is but noting we will need to update two places when needed to be changed

snooty/test_parser.py
```diff
@@ -4327,3 +4328,90 @@ def test_method_selector() -> None:
...
+            <reference refuri="https://www.youtube.com/embed/XrJG994YxD8">
+                <text>https://www.youtube.com/embed/XrJG994YxD8</text>
+            </reference>
+        </directive>
+    </section>
+</root>
+""",
+    )
+
+    # Error cases
```
seungpark, line 4401: minor. seems like the second RST was supposed to be for error cases and the top was for valid (?)

seungpark (approved):
this LGTM! minor nit on defining options twice - not a blocker

seungpark:
oh! and update the PR description for `optional` please :D

snooty/parser.py
rayangler: It would be cool if a mapping of all directives and their available options existed and is reusable based on `rstspec`, but I'm not sure if that behavior exists or not. I think we'll have to re-define them ourselves for now?

snooty/test_parser.py
rayangler: Correct. All of the examples in the top one are all valid in the sense that they'll have valid AST output, minus warnings, but those shouldn't impact the video being rendered. The example in the error case results in the AST not being added at all.

Is there anything in particular I need to change?

seungpark: oh gotcha. existence of `diagnostics` seemed like it should be in error cases but i retract!

snooty/rstspec.toml
seungpark: perfect! 👍

PR #621, opened 2024-09-20
DOP-4915: add display_name property to facets

seungpark:
### Ticket

DOP-4915

### Notes

This is a prereq to the structured data being displayed on the front end. (see this tech spec for TechArticle section that specifies that we need to construct Structured Data based on the `target_platform` property of facets.
These changes extract the `display_name` from the [taxonomy file](https://github.com/mongodb/snooty-parser/blob/main/snooty/taxonomy.toml) and appends them to the produced `page.facets`.

### README updates

### DIFF
--- a/snooty/types.py
+++ b/snooty/types.py
@@ -459,7 +459,7 @@ class ProjectConfig:
                 taxonomy_ref["display_name"]
                 or str(taxonomy_ref["name"]).replace(r"\_", " ").title()
             )
-        except Exception as e:
+        except Exception:
             logger.warning(f"Display name not found for invalid facet {unparsed_facet}")
             return ""
 
--- a/snooty/types.py
+++ b/snooty/types.py
@@ -461,6 +461,7 @@ class ProjectConfig:
             )
         except Exception as e:
             logger.warning(f"Display name not found for invalid facet {unparsed_facet}")
+            return ""
 
     @staticmethod
     def _substitute(
--- a/snooty/types.py
+++ b/snooty/types.py
@@ -445,19 +445,24 @@ class ProjectConfig:
         parent_facets: List[Facet], unparsed_facet: Dict[str, Any]
     ) -> str:
         try:
+            # for version facets, return the version name as is
             if unparsed_facet["category"] == "version":
                 return str(unparsed_facet["value"])
+
+            # start with base taxonomy and iterate through parent facets list
             taxonomy_ref = asdict(taxonomy.TaxonomySpec.get_taxonomy())
             for parent in parent_facets:
                 options_list = taxonomy_ref[parent.category] or []
                 taxonomy_ref = [x for x in options_list if x["name"] == parent.value][0]
+
+            # find target unparsed_facet within taxonomy facet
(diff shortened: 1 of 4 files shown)


snooty/types.py
```diff
@@ -362,12 +365,22 @@ def validate_facets(
...
+    def parse_facet(
+        unparsed_facet: Dict[str, Any], parent_facets: List[Facet]
+    ) -> Facet:
         try:
-            facet = Facet(**unparsed_facet)
+            display_name = ProjectConfig.get_facet_display_name(
+                parent_facets, unparsed_facet
+            )
+            facet = Facet(
+                **unparsed_facet,
```
mmeigs, line 376: TIL

mmeigs (approved):

snooty/types.py
```diff
@@ -427,6 +440,29 @@ def merge_facets(
...
+            options_list = taxonomy_ref[unparsed_facet["category"]]
+            taxonomy_ref = [
+                x for x in options_list if x["name"] == str(unparsed_facet["value"])
+            ][0]
+            return str(
+                taxonomy_ref["display_name"]
+                or str(taxonomy_ref["name"]).replace(r"\_", " ").title()
+            )
+        except Exception:
+            logger.warning(f"Display name not found for invalid facet {unparsed_facet}")
```
rayangler, line 463: (Non-blocking question) Is there a reason why we use `logger` here instead of adding a `diagnostic` to help find the invalid facet? Is it just more convenient to use the logger in this case?

snooty/types.py
```diff
@@ -427,6 +440,29 @@ def merge_facets(
...
+            for parent in parent_facets:
+                options_list = taxonomy_ref[parent.category] or []
+                taxonomy_ref = [x for x in options_list if x["name"] == parent.value][0]
+            options_list = taxonomy_ref[unparsed_facet["category"]]
+            taxonomy_ref = [
+                x for x in options_list if x["name"] == str(unparsed_facet["value"])
+            ][0]
+            return str(
+                taxonomy_ref["display_name"]
+                or str(taxonomy_ref["name"]).replace(r"\_", " ").title()
```
rayangler, line 460: Do we need to account for `\-` for examples like `data-federation`? Is there an example facet that would use `\_`?

rayangler (commented):
Looks cool. Do we know if this change is backwards compatible with how facets are ingested by Mut/Docs Search Transport (i.e. will this change not break anything else that depends on it)? I can also go ahead and test, if needed, but figured I'd ask first just in case

snooty/types.py
seungpark: thought this might be noisy as we already have a check for the [project's facets.toml](https://github.com/mongodb/snooty-parser/blob/main/snooty/types.py#L391) file as well as the [pages facets directives](https://github.com/mongodb/snooty-parser/blob/main/snooty/parser.py#L601)

snooty/types.py
seungpark: good catch! updated the regex

rayangler (approved):

PR #625, opened 2024-10-03
DOP-3424: "On this page" subnav incorrectly displays headings for all language tabs

mayaraman19:
### Ticket

DOP-3424

### Notes

Creates a `selector_ids` object that reflects the AST in terms of the order of tabs/method selectors. This object looks something like this:

```
method-option: abc
children: {
     tab: def
     children: {
          tab: efg
     }
}
```

This is used by snooty ([see PR](https://github.com/mongodb/snooty/pull/1262)) to determine correct headings to show.

### README updates

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -470,7 +470,7 @@ class ContentsHandler(Handler):
         depth: int
         id: str
         title: Sequence[n.InlineNode]
-        selector_ids: object
+        selector_ids: Dict[Any, Any]
 
     def __init__(self, context: Context) -> None:
         super().__init__(context)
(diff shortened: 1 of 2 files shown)


snooty/postprocess.py
```diff
@@ -509,8 +517,13 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
             self.current_depth += 1
             return
 
-        if isinstance(node, n.Directive) and node.name == "method-option":
-            self.scanned_pattern.append(node.options["id"])
+        # add tab selectors here as well
```
seungpark, line 520: minor. reminder to remove comment

snooty/postprocess.py
```diff
@@ -509,8 +517,13 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
...
+
+        if isinstance(node, n.Directive):
+            if node.name == "method-option":
+                self.scanned_pattern.append((node.name, node.options["id"]))
+            elif node.name == "tab":
```
seungpark, line 525: just noting that collapsibles here are not required since we can expand and scroll to headings within collapsibles

snooty/postprocess.py
```diff
@@ -470,15 +470,23 @@ class HeadingData(NamedTuple):
         depth: int
         id: str
         title: Sequence[n.InlineNode]
-        selector_id: Optional[str]
+        selector_ids: Dict[Any, Any]
```
seungpark, line 473: can we type this?

snooty/postprocess.py
```diff
@@ -470,15 +470,23 @@ class HeadingData(NamedTuple):
...
         super().__init__(context)
         self.contents_depth = sys.maxsize
         self.current_depth = 0
         self.has_contents_directive = False
         self.headings: List[ContentsHandler.HeadingData] = []
-        self.scanned_pattern: List[str] = []
+        self.scanned_pattern: List[Tuple[str, str]] = []
+
+    def scan_pattern(self, arr: List[Tuple[str, str]]) -> Dict[Any, Any]:
+        if arr is []:
```
seungpark, line 484: minor - can use `if not arr`. also wondering if preference is to have empty object or None

seungpark (commented):
taking a look at the frontend PR! i think it'll help me understand the usage of `page.ast.options['headings'][].selector_ids`

snooty/postprocess.py
i80and: Note: `is` is the identity operator, so this condition will *always* be false:

```
>>> [] is []
False
```

rayangler (approved):

i80and (approved):

PR #627, opened 2024-10-16
DOP-5060: Number footnotes reference ids based on page context

rayangler:
### Ticket

DOP-5060

### Notes

* This PR allows a postprocess handler to edit footnote reference ID numbers so that footnotes in includes are numbered chronologically within the context of the page properly.
* Please feel free to compare any footnote examples, but here's a few:
  * Test page - [Snooty main branch](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/sandbox/cloud-docs/raymund.rodriguez/main/testing-footnotes/index.html) vs. [staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-5060/cloud-docs/raymund.rodriguez/main/testing-footnotes/index.html) ([rST](https://github.com/rayangler/cloud-docs/blob/sandbox/source/testing-footnotes.txt)) - note the difference between footnotes from includes are linking properly in the staging link compared to the one from the Snooty `main` branch
  * Page from ticket - [Prod](https://www.mongodb.com/docs/atlas/tutorial/create-new-cluster/) vs. [staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-5060/cloud-docs/raymund.rodriguez/main/tutorial/create-new-cluster/index.html)

### README updates


mmeigs (approved):
This looks good, and I love the detailed test.

Not for this PR, but is this supposed to scroll to the footnote reference on initial load? Because mostly it doesn't. I think I got one of them to work if I was originally in the correct tabs.

rayangler:
> is this supposed to scroll to the footnote reference on initial load? Because mostly it doesn't. I think I got one of them to work if I was originally in the correct tabs.

Seems like the behavior is happening on prod as well. If you're seeing the reference hash jumping to the wrong part of the page, this might be happening due to tab content rendering after initial load.

If linking to a footnote reference inside of a tab, it should only work if that tab is active in general at the moment. I don't believe we currently support linking to things inside of tabs on initial load or when they're inactive, so this is most likely expected for now as well

PR #629, opened 2024-11-07
DOP-4945: Default tab available for drivers tabs

mmeigs:
### Ticket

DOP-4945

### Notes

New optional option `:default-tabid:` for `tabs-selector` and `tabs-pillstrip` directives for use with drivers tabs.

Staging links available on sister PR in snooty [here](https://github.com/mongodb/snooty/pull/1301).

### README updates

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -63,6 +63,7 @@ from .diagnostics import (
     SubstitutionRefError,
     TargetNotFound,
     UnexpectedDirectiveOrder,
+    UnknownDefaultTabId,
     UnnamedPage,
     UnsupportedFormat,
 )
@@ -683,6 +684,16 @@ class TabsSelectorHandler(Handler):
 
                 # If default_tabs are present, append to page options
                 if tabset_name in self.default_tabs:
+                    default_tab_is_in_selectors = (
+                        self.default_tabs[tabset_name]
+                        in page.ast.options["selectors"][tabset_name].keys()
+                    )
+                    if not default_tab_is_in_selectors:
+                        self.context.diagnostics[fileid_stack.current].append(
+                            UnknownDefaultTabId(self.default_tabs[tabset_name] or "", 0)
+                        )
+                        return
+
                     if not page.ast.options.get("default_tabs"):
                         page.ast.options["default_tabs"] = {}
 
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -620,7 +620,7 @@ class TabsSelectorHandler(Handler):
                 )
                 return
 
-            if tabset_name == "drivers" and node.options.get("default-tabid"):
+            if tabset_name == "drivers" and "default-tabid" in node.options:
                 self.default_tabs[tabset_name] = node.options.get("default-tabid")
 
             self.selectors[tabset_name] = []
@@ -682,7 +682,7 @@ class TabsSelectorHandler(Handler):
                 }
 
                 # If default_tabs are present, append to page options
-                if self.default_tabs.get(tabset_name):
+                if tabset_name in self.default_tabs:
                     if not page.ast.options.get("default_tabs"):
                         page.ast.options["default_tabs"] = {}
 
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -578,7 +578,7 @@ class ContentsHandler(Handler):
 class TabsSelectorHandler(Handler):
     def __init__(self, context: Context) -> None:
         super().__init__(context)
-        self.default_tabs: Dict[Optional[str], Optional[str]] = {}
+        self.default_tabs: Dict[str, Optional[str]] = {}
         self.selectors: Dict[str, List[Dict[str, MutableSequence[n.Text]]]] = {}
         self.scanned_pattern: List[str] = []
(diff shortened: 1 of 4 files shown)


snooty/postprocess.py
```diff
@@ -677,6 +681,16 @@ def exit_page(self, fileid_stack: FileIdStack, page: Page) -> None:
                     for tabid, title in tabsets[0].items()
                 }
 
+                # If default_tabs are present, append to page options
+                if tabset_name in self.default_tabs:
+                    if not page.ast.options.get("default_tabs"):
+                        page.ast.options["default_tabs"] = {}
+
+                    assert isinstance(page.ast.options["default_tabs"], Dict)
+                    page.ast.options["default_tabs"][tabset_name] = self.default_tabs[
```
seungpark, line 690: curious why we are making this a page option than an option on the resulting directive

snooty/postprocess.py
```diff
@@ -619,6 +620,9 @@ def enter_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
                 )
                 return
 
+            if tabset_name == "drivers" and "default-tabid" in node.options:
```
seungpark, line 623: can we also verify that this is part of the tab ids present

seungpark (commented):
all those back and forth's of `default tab` 🙌

snooty/postprocess.py
rayangler: If I were to guess, it might be for convenience? If a page has multiple instances of `tabs-drivers` found on different parts of the page, then a writer would have to make sure every instance has the same default.

snooty/postprocess.py
```diff
@@ -578,6 +578,7 @@ def exit_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
 class TabsSelectorHandler(Handler):
     def __init__(self, context: Context) -> None:
         super().__init__(context)
+        self.default_tabs: Dict[Optional[str], Optional[str]] = {}
```
rayangler, line 581: Do both the key and value in the dictionary need to be an optional string? I think the key can safely be a `str` for sure. Do we ever want to support the value being`None`? The check for `"default-tabid" in node.options` seems to suggest we always want a string

snooty/postprocess.py
rayangler: Wait, I misread the original question. Thought this was about `tabs`. Ignore me 😅

mmeigs: I was working a bit backwards too, because the tab-context in Snooty uses the page's ast options to decide the default tab and the options available. 

Both tab-selectors and tab-pillstrip are parent directives. Therefore, it's a page-wide directive? Tell me if I'm completely wrong.

seungpark: ah i see, i guess this would be a page option since each (driver) tab on this page should be the same value anyway. i was thinking in the general realm of tabs if we can have a default value option, but for tabs selectors it should be consistent across the page 👍

seungpark (commented):
two minor comments but otherwise LGTM !

snooty/postprocess.py
```diff
@@ -683,6 +684,16 @@ def exit_page(self, fileid_stack: FileIdStack, page: Page) -> None:
...
                 if tabset_name in self.default_tabs:
+                    default_tab_is_in_selectors = (
+                        self.default_tabs[tabset_name]
+                        in page.ast.options["selectors"][tabset_name].keys()
+                    )
+                    if not default_tab_is_in_selectors:
+                        self.context.diagnostics[fileid_stack.current].append(
+                            UnknownDefaultTabId(self.default_tabs[tabset_name] or "", 0)
+                        )
+                        return
```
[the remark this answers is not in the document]
mmeigs: The rest of the assertions all have to do with adding the default_tabs to the page options. But if the default tab id doesn't exist on the page, we don't need to add it to the page options. It's incorrect, right?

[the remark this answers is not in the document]
seungpark: thanks for explaining!

snooty/postprocess.py
```diff
@@ -578,7 +579,7 @@ def exit_node(self, fileid_stack: FileIdStack, node: n.Node) -> None:
 class TabsSelectorHandler(Handler):
     def __init__(self, context: Context) -> None:
         super().__init__(context)
-        self.default_tabs: Dict[Optional[str], Optional[str]] = {}
+        self.default_tabs: Dict[str, Optional[str]] = {}
```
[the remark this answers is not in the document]
seungpark: we could change line 625 to a default value `node.options.get("default-tabid", "")` but very minor here

seungpark (approved):
LGTM ! thanks for explaining above !

snooty/rstspec.toml
```diff
@@ -344,6 +344,12 @@ options.glob = "flag"
 [directive.tabs-selector]
 help = """Add dropdown selector used to select the specified tabset."""
 argument_type = "string"
+options.default-tabid = "string"
+example = """.. tabs-selector:: ${1: string}
+   :default-tabid: ${2:id of the default tab (Optional for drivers)}
+
+   ${0:Content block}
```
rayangler, line 351: We should remove the content block here since the directive doesn't support it

PR #631, opened 2024-11-12
DOP-5135: Remove nested directive error for method-selector

rayangler:
### Ticket

DOP-5135

### Notes

* Removes nested directive error for `method-selector`, as requested by the docs team. Approach was to detach the postprocess handler for `method-selector` from its superclass that handled checking for both nested directives and duplicate directives.
* Removes tests related to instances of `method-selector` nested in other directives.

### README updates

### DIFF
--- a/snooty/postprocess.py
+++ b/snooty/postprocess.py
@@ -2042,7 +2042,6 @@ class MethodSelectorHandler(Handler):
 
     def __init__(self, context: Context) -> None:
         super().__init__(context)
-        self.method_selector_name = "method-selector"
         self.method_option_name = "method-option"
         self.method_description_name = "method-description"
         self.within_description = False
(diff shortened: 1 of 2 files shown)


snooty/postprocess.py
```diff
@@ -2037,23 +2037,26 @@ def __init__(self, context: Context) -> None:
...
 
 
-class MethodSelectorHandler(NestedDirectiveHandler):
+class MethodSelectorHandler(Handler):
     """Handles page-level validations for method-selector directive and its children."""
 
     def __init__(self, context: Context) -> None:
-        super().__init__(context, "method-selector", {"include", "sharedinclude"})
+        super().__init__(context)
+        self.method_selector_name = "method-selector"
```
i80and, line 2045: If this is only read in one place, I would say either make it a global constant or just use the literal directly on line 2075: instance members like this are quite slow to access, and generally less unnecessarily mutable state is better.

i80and (approved):

PR #634, opened 2024-11-19
DOP-5205: Improve onboarding docs

i80and:
### Ticket

DOP-5205

### README updates

- - [x] This PR introduces changes that should be reflected in the README.md and/or HACKING.md, and I have made those updates.
- - [ ] This PR does not introduce changes that should be reflected in the README.md and/or HACKING.md


HACKING.md
```diff
@@ -1,5 +1,7 @@
 # Snooty Parser
 
+![Data flowchart](./docs/flowchart.svg)
```
rayangler, line 3: (Non-blocking) This flowchart talks about adding the AST to the db, but that's now handled by the persistence module. We should consider updating a bit when we have the time!

HACKING.md
```diff
@@ -184,3 +186,29 @@ tag if it was created.
 * Flutter is currently a fork to add support for line numbers. We need to
   figure out a cleaner way of doing this so we can merge it into the
   upstream codebase.
+
+## Glossary
+
+<dl>
+  <dt>reStructuredText</dt>
```
seungpark, line 193: [minor nit] it would be nice to have links to official API documentation for anywhere we are using an external source ie. https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html

HACKING.md
```diff
@@ -184,3 +186,29 @@ tag if it was created.
...
+
+
+  <dt>tinydocutils</dt>
+  <dd>Our vendored fork of docutils, modified to be statically typed and less reliant on method dispatch through string manipulation.</dd>
+
+  <dt>Giza</dt>
+  <dd>The primary entry point of the docs' original tech stack: it would download assets, generate reStructuredText from YAML files (hence the `gizaparser/` directory for compatibility), and invoke Sphinx.</dd>
+
+  <dt>Sphinx</dt>
+  <dd>The primary unofficially official documentation toolchain for reStructuredText.</dd>
```
seungpark, line 213: 😂

HACKING.md
```diff
@@ -184,3 +186,29 @@ tag if it was created.
...
+  <dd>The canonical parsing library for reStructuredText.</dd>
+
+
+  <dt>tinydocutils</dt>
+  <dd>Our vendored fork of docutils, modified to be statically typed and less reliant on method dispatch through string manipulation.</dd>
+
+  <dt>Giza</dt>
+  <dd>The primary entry point of the docs' original tech stack: it would download assets, generate reStructuredText from YAML files (hence the `gizaparser/` directory for compatibility), and invoke Sphinx.</dd>
+
+  <dt>Sphinx</dt>
```
seungpark, line 212: CMIIW but we don't necessarily use sphinx or docutils directly, but a rendition of these, managed in intersphinx and tinydocutils. can we outline that somewhere?

seungpark (approved):
beautiful! thank you for doing this! the diagram being included should help draw a mental image <3

HACKING.md
```diff
@@ -184,3 +186,39 @@ tag if it was created.
...
+  <dd>A markup language identified with the Python ecosystem.</dd>
+
+  <dt>Abstract Syntax Tree (AST)</dt>
+  <dd>A tree of nodes which reflect the syntactic structure of an unparsed textual document.</dd>
+
+  <dt>Postprocessor</dt>
+  <dd>The component of snooty which performs global link analysis and other forms of processing that spans multiple files.</dd>
+
+  <dt>docutils</dt>
+  <dd>The canonical parsing library for reStructuredText. We use our own vendored fork called `tinydocutils`, modified to be statically typed and less reliant on method dispatch through string manipulation.</dd>
```
rayangler, line 203: The backticks for inline code don't seem to be parsed here. Would we need to use `<code>`?

i80and: Ah, drat. Markdoooown!!! *shakes fist* Good catch thanks!

HACKING.md
i80and: We technically support The Good Parts of reStructuredText, not the entire thing. It's a recurring problem that people go to that page, and then try to use misfeatures that should not exist.

Unfortunately, we don't have authoritative documentation on what we do support, but I would rather not direct people to a known bad source.

HACKING.md
rayangler: Nice. I believe `tt` is considered deprecated and `code` is preferred, but this may be more of a nitpick at this point

rayangler (approved):
Nice changes!

HACKING.md
i80and: Never mind, I'm just le silly!

HACKING.md
i80and: Updated!

PR #637, opened 2024-12-10
DOP-5240: Ensure default tabids don't persist across pages

rayangler:
### Ticket

DOP-5240

### Notes

* Default tabid was not being reset whenever a new page was entered. The persisted data from a previous page would then affect the validations and default tabid for a subsequent page, even if it did not have a default tabid set.

### README updates

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [v0.18.8] - 2024-11-22
 
 ### Added
+
 - "mongoid-api" role (#633)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,9 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [v0.18.8] - 2024-11-22
 
 ### Added
+
```
rayangler, line 13: Unrelated to this bug, but is a linting issue on `main` that is causing checks to fail

mmeigs (approved):
Thank you so much

PR #639, opened 2024-12-17
DOP-5250: Consider .ast files as .txt page

rayangler:
### Ticket

DOP-5250

### Notes

* I originally thought that [this line](https://github.com/mongodb/snooty-parser/blob/2c537a2cdee274f957b11c6ed5e0cdf35b10750e/snooty/parser.py#L1881) was enough to treat pages created from `.ast` files to be the same as `.txt` files, but turns out that was wrong. This PR ensures that `.ast` files have a `.txt` fileid to match the parser's, frontend's, persistence module's, etc.'s, assumptions of a page AST having a `.txt` fileid. 
  * The potential con of this approach is that diagnostics for `.ast` files will now include a fileid with a `txt` extension, which might be confusing for anyone who sees `ast`-related issues and are not familiar with the conversion.
* This PR also fixes uncaught bugs that were happening because `.ast` files weren't properly going through the postprocessor. Notably:
  * Missing ids from `n.Heading`
  * `n.ToctreeDirective` always being chosen as the node type for serialized `directive` node objects

### README updates

### DIFF
--- a/snooty/util.py
+++ b/snooty/util.py
@@ -491,7 +491,6 @@ class NodeDeserializer:
         n.Target,
         n.TargetIdentifier,
         n.Text,
-        n.TocTreeDirective,
         n.Transition,
     ]
     node_classes: Dict[str, Type[n.Node]] = {
@@ -512,7 +511,8 @@ class NodeDeserializer:
             if field.name == "span":
                 continue
 
-            node_value = node.get(field.name)
+            placeholder_span = (0,)
+            node_value = node.get(field.name)        
             has_nested_children = field.name == "children" and issubclass(
                 node_type, n.Parent
             )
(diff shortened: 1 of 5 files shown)


snooty/util.py
```diff
@@ -539,12 +545,20 @@ def deserialize(
...
                 # Ideally, we validate that the data types of the fields match the data types of the JSON node,
                 # but that requires a more verbose and time-consuming process. For now, we assume data types are correct.
                 filtered_fields[field.name] = node_value
 
-        return node_type((0,), **filtered_fields)
+        deserialized_node = node_type(placeholder_span, **filtered_fields)
+
+        # Finalize any needed node types and their fields
+        if isinstance(deserialized_node, n.Heading) and not deserialized_node.id:
+            deserialized_node.id = make_html5_id(deserialized_node.get_text().lower())
```
mmeigs, line 559: I believe this also needs the spaces replaced by dashes!

i80and: @mmeigs this is what `make_html5_id()` does, based on the `PAT_INVALID_ID_CHARACTERS` regex `[^\w_\.\-]`!

mmeigs: oh whoops, thanks!

mmeigs (approved):

PR #652, opened 2025-03-28
DOP-5477: ingest composable tutorials from parser

seungpark:
### Ticket

DOP-5477

This PR allows the parser to ingest new directive types `... composable-tutorial` and `.. selected-content::` for a page wide directive.

The parser changes handles the node construction:
- Handles changing  from `@node.options[str]` to `@composableNode.composable_options[dict[]]`
- Handles usage of dictates default values 
- Error handling of wrong ids, missing default ids, missing children, incorrect children directives

The postprocess change handles any usage of full page directive usage (ie. method selectors, tab selectors)

### README updates


.github/workflows/release.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   test:
     strategy:
       matrix:
-        platform: [ubuntu-20.04, macos-13, macos-latest]
+        platform: [ubuntu-22.04, macos-13, macos-latest]
```
mmeigs, line 12: 👏

snooty/n.py
```diff
@@ -512,3 +512,17 @@ class Transition(Node):
...
     __slots__ = ()
     type = "table"
+
+
+@dataclass
+class ComposableDirective(Directive):
+    __slots__ = "composable_options"
+    composable_options: List[
+        Dict[str, Union[str, List[Dict[str, str]], Dict[str, str]]]
+    ]
```
mmeigs, line 522: Referring to this Union: I see examples in the tech spec of the `str`, I see examples of the `List[Dict[str, str]]`, but I don't see examples of the `Dict[str, str]`... Can you point that out? ...also I will probably see this as I go further into the review lol

snooty/parser.py
```diff
@@ -854,6 +862,180 @@ def handle_method_option(self, node: n.Directive) -> None:
...
+        default_ids_dict: Dict[str, str] = {}
+        for index in range(len(option_ids)):
+            option_id = option_ids[index]
+            try:
+                self.check_valid_option_id(
+                    option_id, node, spec_composables_dict, used_ids
+                )
+                composable_from_spec = next(
+                    (option for option in spec_composables if option.id == option_id),
+                )
```
mmeigs, line 904: Maybe we can use a default fallback `None` for this (next intakes a second arg as fallback value) and then have a check right after to send a specific Diagnostic?

snooty/parser.py
```diff
@@ -854,6 +862,180 @@ def handle_method_option(self, node: n.Directive) -> None:
...
+                    option.id: option for option in composable_from_spec.options
+                }
+                self.check_valid_option_id(
+                    specified_default_id, node, allowed_values_dict, set()
+                )
+                default_ids_dict[option_id] = (
+                    specified_default_id
+                    or composable_from_spec.default
+                    or composable_from_spec.options[0].id
+                )
```
mmeigs, line 916: Same here as above: Maybe on 915 we can add ` if composable_from_spec.options else None` or `if composable_from_spec.options[0] else None` (unsure which is best) and then after check `if default_ids_dict[option_id] is None` and raise a specific Diagnostic?

snooty/parser.py
```diff
@@ -854,6 +862,180 @@ def handle_method_option(self, node: n.Directive) -> None:
...
+                    "default": specified_default_id
+                    or composable_from_spec.default
+                    or "",
+                    "dependencies": composable_from_spec.dependencies or {},
+                    "selections": [],
+                }
+                composable_options.append(composable_option)
+
+            except ChildValidationError:
+                continue
```
mmeigs, line 932: Oh but now I see this. Feel free to take on my two comments above or let them flow into this and/or the validation below.

mmeigs (approved):
🌟 Really nice variable naming

snooty/n.py
mmeigs: I saw it later in the review!

snooty/postprocess.py
```diff
@@ -2143,6 +2143,47 @@ def exit_page(self, fileid_stack: FileIdStack, page: Page) -> None:
...
+        self.composable_node_start = node.start[0]
+
+    def exit_page(self, fileid_stack: FileIdStack, page: Page) -> None:
+        if not self.composable_tutorial:
+            return
+        for colliding_ast_option in self.colliding_ast_options:
+            if page.ast.options.get(colliding_ast_option, None):
+                self.context.diagnostics[fileid_stack.current].append(
+                    UnexpectedDirectiveOrder(
+                        f"{self.target_directive_name} cannot be used with {colliding_ast_option} on the same page",
```
rayangler, line 2180: (Nit) I think `["selectors", "has_method_selector"]` are options specific to internal implementation and are not user facing. To make this error more user friendly, would it make sense to print out the associated directive in the error instead (`tabs-selector` and `method-selector`)?

snooty/parser.py
```diff
@@ -854,6 +862,180 @@ def handle_method_option(self, node: n.Directive) -> None:
...
+                    specified_default_id, node, allowed_values_dict, set()
+                )
+                default_ids_dict[option_id] = (
+                    specified_default_id
+                    or composable_from_spec.default
+                    or composable_from_spec.options[0].id
+                )
+
+                composable_option: Dict[
+                    str, str | List[Dict[str, str]] | Dict[str, str]
```
rayangler, line 919: It might be beneficial to have a shared type for: `Dict[str, str | List[Dict[str, str]] | Dict[str, str]` since it's fairly specific and is used in a few different places.

snooty/parser.py
```diff
@@ -854,6 +862,180 @@ def handle_method_option(self, node: n.Directive) -> None:
...
+                )
+
+            except ChildValidationError:
+                continue
+
+        if not default_values_found:
+            self.diagnostics.append(
+                MissingChild(
+                    "composable-tutorial",
+                    "selected-content with selections {default_ids_as_string}",
```
rayangler, line 997: Should this be an f string?

snooty/parser.py
```diff
@@ -854,6 +862,180 @@ def handle_method_option(self, node: n.Directive) -> None:
...
+            )
+        node.composable_options = composable_options
+        node.options = {}
+
+    def handle_composable_content(
+        self,
+        node: n.ComposableContent,
+        spec_composables: List[Composable],
+    ) -> None:
+        selection_ids = re.split(r"\s*,\s*", node.options.get("selections", "").strip())
```
rayangler, line 1009: (Nit) This operation seems to be used in a few places and it's pretty useful for this string pattern. Is there a way to reuse as a function?

snooty/parser.py
```diff
@@ -854,6 +862,180 @@ def handle_method_option(self, node: n.Directive) -> None:
...
+            selection_id = selection_ids[idx]
+            spec_composable = spec_composables[idx]
+            allowed_selection_ids = list(map(lambda x: x.id, spec_composable.options))
+            # check if dependencies are met - then None is not allowed
+            met_dependencies: bool = all(
+                selections[composable_option_id] == option_value_id
+                for composable_option_id, option_value_id in (
+                    spec_composable.dependencies or {}
+                ).items()
+            )
```
rayangler, line 1022: (Question) For my own understanding, why is "None" not allowed in this case?

snooty/parser.py
```diff
@@ -854,6 +862,180 @@ def handle_method_option(self, node: n.Directive) -> None:
...
+                self.diagnostics.append(
+                    UnknownOptionId(
+                        "composable-tutorial",
+                        selection_id,
+                        allowed_selection_ids,
+                        node.start[0],
+                    )
+                )
+                break
+            composable_option_value = str(spec_composable.id)
```
rayangler, line 1033: `spec_composable.id` seems to be of type string. Is this line necessary?

snooty/parser.py
seungpark: All selections should have a choice. None is reserved for when that configurable option does not have its dependencies met.

example:
```
.. composable-tutorial::
   :options: interface, language, cluster-topology, cloud-provider
   :defaults: driver, nodejs, repl, gcp

   .. selected-content::
      :selections: driver, None, repl, aws
```

should result in an error because the `language` field's dependencies has been met

snooty/parser.py
seungpark: good catch!

snooty/parser.py
rayangler: Gotcha, this is a good example. Thanks!

seungpark: adding it to the tests!

snooty/parser.py
seungpark: removed!

snooty/parser.py
seungpark: added it [here](https://github.com/mongodb/snooty-parser/pull/652/files#diff-e43a5722b09dcc8e1ea71a8a0104599496fba59b067d05ae7a7f1cb5028b6534R4698-R4707) (last case of `test_composable_tutorial_errors` )

snooty/postprocess.py
seungpark: updated! this is a bit hardcoded at this handler as a tuple, but feel free if you have other suggestions here

snooty/parser.py
seungpark: i did it for best practice, but essentially should not be hitting that line since we are already checking beforehand in the `check_valid_option_id` calls. also noting the tests didnt double down on the errors for invalid options :)

mmeigs (approved):

snooty/parser.py
rayangler: Makes sense. I don't have too much of a preference on the reduced form, but as long as it works for our use case, I'm okay with it!

snooty/postprocess.py
rayangler: I think this is fine for now since it's a pretty small list.

If the list grows/changes in the future, we can maybe have the `enter_node` method look for and keep track if one of those directives were seen on the same page.

rayangler (approved):

PR #664, opened 2025-05-20
DOP-5604: dismissible skills

mmeigs:
### Ticket

DOP-5604

### Notes

Creates new `dismissible-skills-card` directive. The subsequently created node is injected into the page options as it is not in the normal page flow on the frontend.

The only necessary information needed is the skill title and the url to the skill.

A PR for Snooty is being worked on - will link here when open.

### README updates


seungpark (approved):
this LGTM. can we add a postprocess test here to validate the output

rayangler (approved):

PR #671, opened 2025-08-06
DOP-6049 adds new directives called multi-column and column

caesarbell:
### Ticket

DOP-6049

### Notes

Adds two new directives (multi-column, and column).

### README updates

### DIFF
--- a/snooty/rstspec.toml
+++ b/snooty/rstspec.toml
@@ -586,7 +586,7 @@ content_type = "block"
 options.title = "string"
 
 [directive."multi-column"]
-help = """A two-column component for the homepage"""
+help = """A multi-column component for the homepage"""
 content_type = "block"
 example = """.. multi-column::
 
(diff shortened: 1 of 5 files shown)


snooty/rstspec.toml
```diff
@@ -580,6 +580,32 @@ help = """A list of languages that this guide supports."""
 # """
 content_type = "list"
 
+[directive.column]
+help = """column content for the n-column component"""
+content_type = "block"
+options.title = "string"
+
+[directive."multi-column"]
+help = """A two-column component for the homepage"""
```
seungpark, line 589: change to multi column component

seungpark: also, would add a note (or option) to note how each column would be presented in small vs big screen. the design shows each column being its own block

seungpark (approved):

